PR #4, opened 2020-08-03
Update to FetchContent

nathangibson14:
Replaced metaconfigured-based build system with FetchContent-based build system.  Ensured code still compiles and all tests pass.  Updates to build system expected to continue, but this puts the basic functionality into place and allows this code to be used in other modules.


staleyLANL:
Thanks for doing this! I'm looking at it right now, to ensure that the new build system works for me as well.

staleyLANL:
Works for me.

PR #8, opened 2020-09-08
Feature/more tests

staleyLANL:
For this PR, I've refurbished GNDStk's test suite.
Many existing test codes have been improved.
I've improved the use of the Catch system, based on some comments that were put in another recent GNDStk PR.
Most (not quite all) previously stubbed test codes now have real content in them.
A couple stub test codes remain; I'll do those in the future, as time allows.


src/GNDStk/JSON/test/JSON.test.cpp
```diff
@@ -4,6 +4,317 @@
...
 
+using namespace njoy::GNDStk;
+
+
+// -----------------------------------------------------------------------------
+// JSON literal
+// -----------------------------------------------------------------------------
+
+static const std::string string_real_json =
+R"***({
```
jlconlin, line 15: Don't you just love raw-string literals

src/GNDStk/JSON/test/JSON.test.cpp
```diff
@@ -4,6 +4,317 @@
...
+      const JSON j("n-069_Tm_170-covar.json");
+      std::ostringstream oss;
+      j.write(oss);
+      THEN("We can check that the result is what we expect") {
+         CHECK(oss.str() == string_real_json);
+      }
+   }
+
+   // to file
+   // fixme How do we handle *writing* a file with this test system?
```
jlconlin, line 290: We write functionality to write to a `std::ostream` or `std::string` object and test that functionality. Writing to a file after you have a string is trivial. ENDFtk doesn't ever write to a file. You just give it an iterator to a string and it starts writing there.

src/GNDStk/Node/test/all.test.cpp
```diff
@@ -18,8 +18,10 @@ SCENARIO("Testing GNDStk Node all()") {
          // nodes: chemicalElement
          bool found = false;
          auto elem = elems.all("chemicalElement",found);
```
jlconlin, line 20: Is passing `found` here required? This feels a lot like C syntax, which makes me sad.

src/GNDStk/Tree/test/write.test.cpp
```diff
@@ -231,43 +219,45 @@ SCENARIO("Testing GNDStk tree write() and operator<<") {
       // write *files*.
 
       // format::tree
-      WHEN("We write() or << the empty tree in format::tree format") {
+      WHEN("We write() the empty tree using format::tree") {
+         // write()
          {
```
jlconlin, line 224: Could make these blocks using `THEN` to help clarify what is being done.

src/GNDStk/child_t/test/operator.test.cpp
```diff
@@ -0,0 +1,227 @@
...
+// SCENARIO
+// ------------------------
+
+SCENARIO("Testing GNDStk child_t operators") {
+
+   // ------------------------
+   // -child_t
+   // ------------------------
+
+   GIVEN("A child_t<type,one|all,converter>") {
```
jlconlin, line 111: So these test the functionality just fine, but they don't help me to understand how to use the library. I don't want you to change this, but in the future, please construct examples at least something closer to what we will actually do with GNDS

src/GNDStk/child_t/src/operator.hpp
```diff
@@ -100,11 +90,29 @@ inline auto operator--(const child_t<TYPE,FIND,CONVERTER> &kwd, const int)
    );
 }
 
-// child_t<TYPE>++
+// operator++
```
jlconlin, line 93: I'm not sure what these operators do.

src/GNDStk/test/GNDStk.test.cpp
```diff
@@ -2,7 +2,3 @@
 #define CATCH_CONFIG_MAIN
 
 #include "catch.hpp"
```
jlconlin, line 4: Was this file intended to be (mostly) empty?

src/GNDStk/type2node.hpp
```diff
@@ -31,6 +31,7 @@ and converting them into a node that's destined to be added as a child node.
    ) { \
       try { \
          node.clear(); \
+         node.name = "values"; \
```
jlconlin, line 34: Why are you setting the node name to `values`?

jlconlin (changes requested):

jlconlin:
Overall, lot's of good cleanup. Have a few clarifications about what is being done.

src/GNDStk/JSON/test/JSON.test.cpp
staleyLANL: It's certainly convenient for users, however, to provide an API that allows direct file writing. I provide functionality to do that, as well as to write to `ostream`. Then, the file writing capability actually just uses an `ofstream` and calls the `ostream` version. In my opinion, one would still ideally test the writing capabilities to write directly to a file, but if it's fine not to (I know it works), that that's good as well. I can just delete the "fixme" you referenced. Deleting fixmes, without having to do any work, isn't something I'll object to. :-)

src/GNDStk/Node/test/all.test.cpp
staleyLANL: Don't let it make you sad. GNDStk's `found` out-parameter is always optional. If you don't provide it, the behavior is like that of other codes in NJOY21: throw an exception if there's a problem. If you do send `found`, then the behavior is a bit different: the function will set `found` to `true` or `false` as appropriate, but *not* throw an error if it wasn't found. This would be convenient for a user who knows that something might or might not be found, is prepared to handle either case, and doesn't want the extra clutter of putting `try/catch` material in their own code.

src/GNDStk/Tree/test/write.test.cpp
staleyLANL: Sure, I can go in and do that. The `THEN` messages throughout this file will be somewhat repetitive, basically saying something along the lines of "then we'll get the empty string" or "we'll get a declaration-node-only XML" or similar verbiage, but I can certainly do that.

src/GNDStk/child_t/test/operator.test.cpp
staleyLANL: The referenced material is from a brand-new test code, that tests some obscure features that I put in mainly because I anticipate possible interest from future developers, perhaps more so than from end users. This test *is* basically made to just mechanically test the features. In a real code, I'd expect somebody to use a mixed bag of features from here and there - all over the code base. Really, I think the bigger discussion here would be about a suite of tests, versus a series of examples (say, in a tutorial) geared towards users. Tests and examples are typically handled differently, even if perhaps in some ideal world they could be one and the same.

src/GNDStk/child_t/src/operator.hpp
staleyLANL: These, as well as the division operators, comprise (part of, for now) what I believe might prove to be a handy, albeit somewhat stylized and probably unexpected, "keyword arithmetic" system I have in mind. I thought you might ask about it.

Without going into too much detail here... Recall the examples I've shown (when I first described it, and also during my WPEC talk) about using what I've called my keyword system: the `child_t` and `meta_t` types that embed information, in particular something's name in GNDS, and a *type* with which you want it to interact.

Basically, for `child_t` and `meta_t`, `operator/` as well as `++` and `--` are designed so that you can take an already-made `child_t` or `meta_t` and easily create another similar one - based mostly on the original one, but with a small modification. Just as one example: let's say you have a `child_t` called `mydate` that's encoded to give you some custom type `date_t`. Then you decide you also want to allow for another format, `date2_t`. Write `auto mydate2 = date2_t{}/mydate`, and it does just that: duplicate everything from `mydate`, except change `<date_t>` somewhere to `<date2_t>`.

Remember the `one` versus `all` business? The `--` and `++` operators, applied to a `child_t`, "upgrade" to `all` or "downgrade" to `one`. (In doing so they actually create a new `child_t`, they don't modify the existing one as `int++` modifies the int in question. The language does allow this.)

In short, it's a rather stylized concept, but one I think will be helpful. It'll especially be helpful if and when I expand the capabilities of `child_t` and `meta_t` - which I plan to do. (@jlconlin in fact you yourself suggested when we first spoke that `child_t` could be outfitted with a "filter" or sorts, so that it would only pull out child nodes, of the given name, that satisfied some user-defined condition. So, imagine going to the syntactic effort to set up such an animal. Then, my keyword-arithmetic system will allow you to easily made other `child_t` objects from the first one, with less effort, merely with some expression like what I wrote a couple of paragraphs above.)

src/GNDStk/test/GNDStk.test.cpp
staleyLANL: For that one, I actually just looked to see what ENDFtk did with the analogous file. It did exactly this. :-) I guess the question is whether or not we need to "test" the main, GNDStk.hpp header file that everyone includes. Maybe we should just delete this??

src/GNDStk/type2node.hpp
staleyLANL: The GNDS content that's relevant to this code typically appears (or, for output, should appear) in a `<values>` node. Depending on just what somebody does to create a new node that they intend to write somewhere, the name might (and usually is) set elsewhere. And it *was* being so set - until I just recently wrote a direct test of this function. At that time, I realized that it would probably be a good idea to have the function put in something, not just `""`, for the default name.

src/GNDStk/child_t/test/operator.test.cpp
jlconlin: I agree. Having a series of examples is a very good idea. Something that we need for people to start using this library. It would be easy to develop the series of examples; anytime someone asks how to do something, we add another example. Can be developed over time.

src/GNDStk/test/GNDStk.test.cpp
jlconlin: If there is something that can only be tested at the highest level, then it should go here. If there is nothing, then let's delete the file—and perhaps even the `test` directory.

jlconlin (approved):

PR #9, opened 2020-09-25
Substantially extended node::operator()'s capabilities.

staleyLANL:
Split node's call.hpp into two files for increased clarity.
Specifically, (meta_t,...) and (child_t,...) calls are split.
Also, constness for the latter is handled with less code.
Slightly reworked child_t one/all downgrade/upgrade syntax.
New "filter" capability for node:: one, all, child, operator().
Small code changes here and there, for clarity and consistency.
Some slight code shortening here and there as well.
Modest improved comments and terminology in a few places.


staleyLANL:
I need to add formal tests for the new material, but thought I'd get this out there.
Summary: pursuing some ideas I'd thought about when writing my L3 talk, I made substantial improvements in the query capabilities. Some of it turned out to be trickier than I'd anticipated, but I think it all turned out quite well.

jlconlin:
I won't be looking at this for the next week. I hope it isn't urgent.

staleyLANL:
No worries. You could always just approve it, because it's fabulous. Or, we could simply wait until I add formal tests, as I'll be doing that before too long anyway.

staleyLANL:
@jlconlin With my latest commit from a little while ago, I think this PR is ready for prime time.

src/GNDStk/Node/test/all.test.cpp
```diff
@@ -1,26 +1,96 @@
...
+
+         // filter that requires: node has metadata ENDF_MT="600"
+         auto endf_mt_600 = [](const node &n)
+            { return n(ENDF_MT) == 600; };
+
+         // filter that requires: both of the above
+         auto filter_both = [nplus,endf_mt_600](const node &n)
+            { return nplus(n) && endf_mt_600(n); };
+
+         THEN("There are 15 <reaction> nodes with label=\"n + *\"") {
```
jlconlin, line 52: I like this idea of a filter. I think it will be very helpful.

src/GNDStk/Node/test/all.test.cpp
```diff
@@ -1,26 +1,96 @@
...
+         }
+      }
+
+      // reactions / reaction
+      WHEN("We extract selected <reactions> nodes within <reaction>") {
+         const auto &Rs = tree(reactionSuite,reactions);
+
+         // filter that requires: node has metadata label="n + *"
+         auto nplus = [](const node &n)
+            { return 0 == strncmp(n(label).c_str(), "n + ", 4); };
```
jlconlin, line 42: I'm not sure what `label` is here. It looks like a variable, but I can't seem to figure out what it is.

src/GNDStk/Node/test/all.test.cpp
```diff
@@ -1,26 +1,96 @@
...
+            CHECK(Rs.all("reaction",filter_both,found).size() == 0);
+            CHECK(!found);
+         }
+      }
+
+      // reactions / reaction
+      WHEN("We invoke node::all() with an empty (\"\") key") {
+         auto Rs = tree(reactionSuite,reactions);
+
+         THEN("We get back a container with one node, identical to *this") {
```
jlconlin, line 72: What is the use case for this? It seems like we are getting a copy of a node. Why not just use the copy constructor?

jlconlin (commented):
I'm intrigued by the new filter capabilities presented here. I think they are going to be really helpful.

jlconlin:
We probably need to spend some time on documentation. There are a lot of features in GNDStk that only @staleyLANL knows about.

src/GNDStk/Node/test/all.test.cpp
staleyLANL: Definitely. I anticipate that a lot of other capabilities will be built upon this. There's a bit more I have in mind to do regarding filters, but there are probably other priorities at the moment.

src/GNDStk/Node/test/all.test.cpp
staleyLANL: It's in GNDStk. Basically,

```
const meta_t<std::string> label("label");
```

So, `label` means: look in the node (in this case called `n`), for the metadatum key/value pair with a key of `"label"`, and return the key's associated value as a `std::string`.

src/GNDStk/Node/test/all.test.cpp
staleyLANL: This is basically a mechanical test of a certain behavior that doesn't appear at first glance to be useful. However, a certain other behavior - one that's quite useful - is built on top of this. I'll have to write a much longer explanation in order to establish why I'm doing this. At the moment I have some other things to do, and will write the long explanation when I get the chance. However, rest assured that we want this.

staleyLANL:
> We probably need to spend some time on documentation. There are a lot of features in GNDStk that only @staleyLANL knows about.

Definitely. We (probably including more than just you and I) should talk about what's best to do next for the project. Off the top of my head, I can think of several directions to go... (1) Documentation. (2) Studying the GNDS specs (in particular those JSONs) and using them to help put together more GNDS-specific data structures. (3) Some additional features I have in mind. (4) Some GNDStk-wide naming issues that I've been thinking about, and that would be wise to do before too many people are using GNDStk. (5) Writing a tutorial. Maybe the last one should be grouped with Documentation. In any event, lots to think about.

staleyLANL:
> I'm intrigued by the new filter capabilities presented here. I think they are going to be really helpful.

Agreed. There's more I want to do in that direction, but I think it should wait for another PR, so we can get this one closed out.

jlconlin (approved):

PR #12, opened 2020-10-20
Handling conversion between node and container differently.

staleyLANL:
Formulated a slick variable template for dealing with <values> nodes.
See type2node.hpp for a comment that sketches the reasoning for this.
The change cascaded into several other changes here and there.
Added more Node assignment tests.
Consolidated some namespace blocks.
Modest rearranging of some code and constructs.
The rearrangement allowed a couple of "detail" files to be deleted.
C++ is not Python. Use ! and && and ||, not not and and and or.
As usual, tweaked comment verbiage and bits of code here and there.


nathangibson14 (commented):
I'm not intimately familiar with the operations you are refactoring here, but I didn't see anything that looked awry.  I'll let another reviewer way in to click the "Approve" button, as I don't know that I provided much value added with my review.

Regarding "or" vs "||", etc.: Aren't both in the C++ standard now?  Is there a reason you felt the need to change course on which one you were using?  I've never seen a strong reason for using one over the other, so I don't object to changing them, I'm just curious what caused you to put the effort in to change them.

src/GNDStk/Tree/src/read.hpp
```diff
@@ -70,7 +70,7 @@ std::istream &read(
          // assume .xml
          // go through a temporary XML object to create the tree...
          const XML tmp(is);
-         if (not is.fail())
+         if (!is.fail())
```
jlconlin, line 73: These changes were not necessary.

src/GNDStk/type2node.hpp
```diff
@@ -9,37 +9,46 @@ The description of those, seen in reverse, applies here.
...
-         throw; \
-      } \
-   }
-
-   GNDSTK_CONVERT(deque)
-   GNDSTK_CONVERT(list)
-   GNDSTK_CONVERT(vector)
-
-#undef GNDSTK_CONVERT
+/*
```
jlconlin, line 24: Comments about what used to be in the code don't need to be there. Now, I have to maintain these comments, which will immediately be out of date when the Pull Request is finished. 

I appreciate the documentation about why this was removed, but it should go in the description of the Pull Request and not in the code itself.

src/GNDStk/type2node/test/type2node.test.cpp
```diff
@@ -19,10 +19,16 @@ SCENARIO("Testing GNDStk convert(type,node)") {
...
+   // Update, 2020-10-19. The relevant convert() functionality has been
+   // reformulated into the detail::convert_pcdata_text_t callable object,
+   // so we'll reformulate with that. Ultimately, due to the location of
+   // the new code, we should probably put this test material elsewhere.
+
    WHEN("type is vector") {
       std::vector<double> container = { 1, 2.3, 4.56 };
       node n("old name");
-      convert(container,n);
+      // convert(container,n);
```
jlconlin, line 30: If you want to have such a comment like this, then you could (not necessarily should) put it in a 

```
THEN( "..." ){
  ...
}
```

block.

src/GNDStk/mixed/src/detail.hpp
```diff
@@ -0,0 +1,160 @@
+
+namespace detail {
+
+// text_metadatum_to_string
+// Helper for cdata and comment
+class text_metadatum_to_string {
```
jlconlin, line 6: Where is this stuff tested?

src/GNDStk/Tree/src/read.hpp
whaeck: Since these changes do not alter the intent of the code, this is a non-issue in my opinion so there's no need to call it out.

src/GNDStk/node2type/test/node2type.test.cpp
```diff
@@ -45,9 +45,15 @@ SCENARIO("Testing GNDStk convert(node,type)") {
...
 
+      // Update, 2020-10-19. The relevant convert() functionality has been
+      // reformulated into the detail::convert_pcdata_text_t callable object,
+      // so we'll reformulate with that. Ultimately, due to the location of
+      // the new code, we should probably put this test material elsewhere.
+
       THEN("convert(foo,deque) extracts the content correctly") {
          std::deque<int> container;
-         convert(foo,container);
+         // convert(foo,container);
```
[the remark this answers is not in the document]
staleyLANL: No worries. The refactoring (really, more than just that) was for a good reason, but one that would require a somewhat long explanation here. Suffice it to say that there was something that would have done the right thing most of the time, by default, but been more awkward for users to circumvent in the instances where users might wish to customize something and wanted that particular aspect of it to work differently.

I replaced this "usually right, but awkward to work around when wrong" with, roughly speaking, "undefined by default, but easier to work around; and still defined in the constructs GNDStk provides when you don't customize." Or something like that. If a user writes something custom without providing a function that's needed, they'll get something like "error: no such function" instead of some multi-page convoluted nonsense that ultimately means that my default does the wrong thing in their case.

Months ago someone suggested that I use the keywords: `or`, `and`, etc. They *are* standard now. However, I've done some checking around, and for certain reasons (for example, creating possibly confusion with languages that use the words but with precedence being different), they're generally considered to be inferior to the original way of doing them, which are clear to any C++ expert. So, I made a decision to change them back.

Ya, I'll remove the commented code. I guess I thought it might be worthwhile for someone to see what the new lines were, effectively, doing. But really, it just doesn't matter.

src/GNDStk/type2node/test/type2node.test.cpp
staleyLANL: I've just gone ahead and removed those.

src/GNDStk/type2node.hpp
staleyLANL: I thought about this at the time, but decided to try to formulate the comment in such a way that it would provide a bit more value than just, "we once had these, but no longer."

Conversions between objects in their tree form, and custom objects, are an integral part of GNDStk. I anticipate that we *may* thus have code again in this file someday (now, it's just comments). That, first, is why I kept the file -- as essentially a placeholder.

And, then, given that I kept the file, I wanted to say something about why I *don't* have something that might seem natural to have -- namely, those particular conversions that I removed.

This way, we'll avoid the possibility that somebody might see the file, think "hey, the original author missed this thing that's so natural to have," and then waste time rewriting essentially the same thing that I intentionally removed, before they finally discover the hard way *why* I removed it.

Someone working on the code in the future is more likely to see what's written directly in the code, right in front of their eyes, than something buried in old PR comments from months or years earlier. That's my thought at least.

I supposed I could reword the comment so that it doesn't even say such-and-such content was *once* there, but just says why it isn't there...period.

staleyLANL: Just taking my previous remark a bit further.... :-) *Good* codes have comments about what they do, and perhaps why. *Great* codes also have, at least in select places, comments about why they *don't* do certain things.

On more than one occasion, I've revisited my own code that I hadn't worked on for years, and thought that something obvious was missing, or that I could do this or that in a simpler way than I was doing it.

Then, after writing code again and running into problems, something in my brain would click, and I'd remember that I dealt with the same issue years before, but had forgotten about it.

So, these days, when I rediscover something in this way (or when I anticipate that I might, as I did here), I write a comment. A comment about what I *don't* have. That'll save someone, possibly myself, some amount of grief down the road. :-)

Very simple example: the comment that begins with "Otherwise..." a few lines up from the bottom of `Node/src/call-child.hpp`. (Although that's not so much about something I don't have, but about why I don't do something that's simpler but in fact wrong.)

src/GNDStk/mixed/src/detail.hpp
staleyLANL: In the current invocation of the test suite, it looks like that particular detail material, `text_metadatum_to_string`, indeed wasn't being tested, even indirectly. So, I put in a test (indirect, but the function does get called now).

Other functions in that file do, in fact, get tested indirectly. I have it in mind to actually change some of that functionality works, and will put that in another PR sometime. Because of that, it probably isn't worthwhile at this time to write direct tests that'll just get tossed after I change the constructs in question.

staleyLANL:
Some changes per the PR comments should be uploaded now.

staleyLANL:
I have a boatload of new york that I'd like to place into another PR. Anything critical that the current one can't do without, or can we get it in? @jlconlin @whaeck @nathangibson14

staleyLANL:
I mean new work, not new york

jlconlin (approved):

jlconlin:
> I have a boatload of new york that I'd like to place into another PR. Anything critical that the current one can't do without, or can we get it in? @jlconlin @whaeck @nathangibson14

If you already have the work, put it in a new Pull Request. Then, let's start working on documentation. Without the documentation, you'll be the only one that can use this.

staleyLANL:
It says I'm not authorized to merge this PR. Even though it's approved. Maybe some permissions could be changed so that I can merge approved PRs?

jlconlin:
Sorry @staleyLANL GitHub permissions are kind of clunky. You should be able to merge GNDStk-related stuff now.

PR #14, opened 2020-10-29
Added some specialty member functions to Node.

staleyLANL:
They're for extracting data from particular types of nodes.
They are: `doc[umentation]()`, `cdata()`, `pcdata()`, `comment[s]()`.
Convert for `pair` supported; seems like it'll be useful for various things.
`convert(istream,container)` stays in `convert()` (not `>>`) for elements.
Put in several tests for the `values<>` variable template we'd built already.
Reworked some detail namespace functionality as necessary.
In particular, for `values_type` traits class.
Also support `values<void>` ==> `Node`.
Better (and, really, correct) use of `found` in `Node.operator()`.


src/GNDStk/Node/src/child.hpp
```diff
@@ -2,27 +2,27 @@
 // -----------------------------------------------------------------------------
 // Node.child(child_t<*>)
 //
-// FIND == find::one
```
jlconlin, line 5: I think `allow` is a much better word than `find`.

src/GNDStk/Node/test/call.test.cpp
```diff
@@ -54,12 +54,284 @@ struct isotope_t {
...
 }
 
 
+// -----------------------------------------------------------------------------
+// values_string
+// -----------------------------------------------------------------------------
+
+static const std::string values_string =
+R"***(1e-05
+460000
```
jlconlin, line 69: Can you reduce the number of numbers here? I don't think we need all of these to adequately test these capabilities.

src/GNDStk/misc/src/child.hpp
```diff
@@ -0,0 +1,345 @@
...
+// Keywords
+// Of type child_t
+// -----------------------------------------------------------------------------
+
+// ------------------------
+// one + many pair
+// constructs
+// ------------------------
+
+GNDSTK_MAKE_CHILD(void, averageEnergies, allow::one);
```
jlconlin, line 28: This file looks like something that could be automated using the JSON files that describe the GNDS format.

jlconlin (commented):
Mostly find/replace changes, but others are more significant.

Requesting a few changes. Still need to look at a few files.

staleyLANL:
> Mostly find/replace changes, but others are more significant.
> 
> Requesting a few changes. Still need to look at a few files.

Yeah, the recent stuff is mostly find/replace. But this builds on a ton of earlier stuff. In fact this includes material from the earlier PR, and from this one in its original form, which had some good new stuff. The PR has grown beyond its original purpose, I suppose, but work is work. :-)

src/GNDStk/misc/src/child.hpp
staleyLANL: It may eventually be just that. Or, I may keep these objects in their current form, and have other namespaces to contain objects for, say, version-specific GNDS contents. Lots of ways to handle this sort of stuff.

src/GNDStk/Node/test/call.test.cpp
staleyLANL: Yes, no problem. I'd already reduced it from the full list of values in the file, but can reduce it further. Perhaps by about a factor of two?

src/GNDStk/Node/src/child.hpp
staleyLANL: Agreed. Also, "find" suggested that someone was finding content - i.e. pulling data from a GNDS tree. So, the terminology was biased towards input. "Allow", plus the name change from "all" to "many", now means: Does GNDS allow one such node under a given parent node, or many such nodes? Much more natural terminology, and I/O-agnostic.

src/GNDStk/Node/test/many.test.cpp
```diff
@@ -18,14 +17,15 @@ SCENARIO("Testing GNDStk Node all()") {
...
+            CHECK(CEs.has_many("chemicalElement"));
+            auto ce = CEs.many("chemicalElement");
             THEN("We find eight of them") {
                CHECK(ce.size() == 8);
             }
          } {
             // with "found"
             bool found = false;
-            auto ce = CEs.all("chemicalElement",found);
+            auto ce = CEs.many("chemicalElement",found);
```
whaeck, line 28: I know one(...) will return a reference to the node that was found. Does many(...) return a container of references, or a container of copies of the nodes that were found?

src/GNDStk/keyword/test/keyword.test.cpp
```diff
@@ -160,24 +159,21 @@ SCENARIO("Testing GNDStk keyword") {
...
+      CHECK(tree(xml).children.size() == 0);
+      CHECK(tree(xml,version) == "1.0");
+      CHECK(tree(xml,encoding) == "UTF-8");
       CHECK(tree(covarianceSuite).metadata.size() == 4);
       CHECK(tree(covarianceSuite).children.size() == 3);
       CHECK(tree(covarianceSuite,projectile) == "n");
       CHECK(tree(covarianceSuite,target) == "Tm170");
       CHECK(tree(covarianceSuite,evaluation) == "ENDF/B-8.0");
-      CHECK(tree(covarianceSuite,mixed::meta::format) == 1.9);
+      CHECK(tree(covarianceSuite,format) == 1.9);
```
whaeck, line 171: I'm not sure we should have a default type of double for the format version of the GNDS file. I'd leave it as a string to be honest. BTW: this is merely an observation, no need to change it right now.

whaeck (approved):
Quite a few cosmetic changes (changing find to allow) and some additional features (the has(...) capability, regex on keywords, some additional operator overloading for the call operator on a node, etc.).

You actually took into account the changes requested by @jlconlin so I think this is good to go.

src/GNDStk/Node.hpp
```diff
@@ -55,13 +55,23 @@ class Node {
...
    #include "GNDStk/Node/src/one.hpp"
-   #include "GNDStk/Node/src/all.hpp"
+   #include "GNDStk/Node/src/many.hpp"
    #include "GNDStk/Node/src/child.hpp"
 
-   #include "GNDStk/Node/src/sort.hpp"
+   // has*
+   // Boolean query: will the analogous access function succeed?
+   // Example: has(A,B,C,D) == "operator()(A,B,C,D) succeeds"
+   #include "GNDStk/Node/src/has.hpp"
```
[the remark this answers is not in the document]
staleyLANL: And I've made a note to write the extension we talked about.

src/GNDStk/Node/test/many.test.cpp
staleyLANL: It returns a container of copies. I'm not sure whether or not we'd actually want a container of references. If we did, I'd of course need to use something like `std::reference_wrapper`.

src/GNDStk/keyword/test/keyword.test.cpp
staleyLANL: Yes, I think you're right. The `mixed` (now `misc`) namespace was meant to contain meta_t and child_t objects equipped here and there with types that seemed reasonable, given what I saw in existing GNDS files. It always was, and it remains, basically an approximation of what might be reasonable types for different objects. It's good for testing, but will probably be deprecated eventually, as we develop (or auto-build from the JSON specs) more-official types.

src/GNDStk/Node/test/many.test.cpp
staleyLANL: Thinking about it, containers of `std::reference_wrapper`s may well make sense, for when users want to modify GNDS trees rather than just query them. This is probably a discussion we should have sometime soon.

whaeck: Yes, using reference wrappers would be nice but it is not urgent for now.

src/GNDStk/Node.hpp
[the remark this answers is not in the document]
whaeck: Thanks ;-)

src/GNDStk/Node/test/many.test.cpp
staleyLANL: Actually, if and when we're sure that containers of reference wrappers wouldn't have funny (in a bad way) semantics, this approach might resolve some questions I've been running into in designing a new node operator[] that's equipped for modifying existing nodes. So I suspect this issue will be revisited sooner rather than later, but probably not immediately.

staleyLANL:
@jlconlin are you good to go with this? It looks like I have merge permissions now, but still want to double check.

staleyLANL:
I'll go ahead and merge, to get this off the table. Additional code comments can always be brought up later.

PR #15, opened 2020-11-19
Feature/node bracket

staleyLANL:
Sorry, slamming y'all with another PR. See my long winded code comments for an explanation of what I'm doing. @whaeck may be able to use this material right away. Note that I haven't formally tested it yet, but I don't think it will brick your computer or anything.

### DIFF
--- a/src/GNDStk/Node/src/subscript.hpp
+++ b/src/GNDStk/Node/src/subscript.hpp
@@ -1,0 +1,275 @@
+
+// -----------------------------------------------------------------------------
+// DISCUSSION
+// -----------------------------------------------------------------------------
+
+/*
+Node's bracket operators, operator[], complement its call operators, but with
+some crucial differences.
+
+The generalized call operators, operator(), allow for any number of arguments,
+where the arguments (generally child_t and possibly meta_t at the end, with some
+caveats) represent a path down which to follow, in the node, to get to some
+final child node or metadatum that should be retrieved. Unless the final value
+is retrieved with a <void> child_t or meta_t - meaning raw form - a conversion
+to some stipulated type is applied. In the latter case the result is necessarily
+returned by value, not by reference, and thus cannot be used as an lvalue.
+
+So, for example, for child_t A, B, and C, and meta_t<int> D, we *cannot* write:
+
+   // no! :-(
+   node(A,B,C,D) = 10; // compiler error
+
+because the left-hand side isn't an lvalue. It is, rather, a returned-by-value
+int that became an int in the first place because D, being a meta_t<int>, caused
+the internal raw std::string metadatum value (that's paired with node(A,B,C)'s
+D metadata key) to be converted to an int.
+
+The above then raises a question: how can we *set* some metadatum or child node,
+as we clearly intend to above?
+
+The answer: use [], our bracket operator. First, however, realize that C++ does
+not, at present, allow operator[] to have multiple arguments. So, to use it, we
+must split off the "last argument":
+
+   // yes! :-)
+   node(A,B,C)[D] = 10; // yay, does what we want!
+
+That's easy, and of course if we'd dug down to the A,B,C node already, it means
+just a plain old [] call:
+
+   auto abc = node(A,B,C);
+   abc[D] = 10; // <== like this
+
+One might ask: If we can trick out the [] operator to produce an lvalue, so as
+to allow for the assignment, then why didn't we just do the same thing for the
+call operator? That is, couldn't operator() have gotten the same treatment we
+gave to operator[], so that having the latter isn't necessary?
+
+The answer is the infamous "yes, BUT" - and that's a big "but." Consider our
+bracket-based construct,
+
+   node(A,B,C)[D]
+
+or equivalently abc[D] as in the second working example. For this to be usable
+as an lvalue, and with the effect we want to achieve, it can't simply be an int,
+like the superficially similar "node(A,B,C,D)" construct. Rather, it must be
+some sort of "smart object" that retains information about (1) the location of
+the internal raw std::string metadatum value that we wish to change; and (2) D's
+converter from int to std::string. Armed with this information, the smart object
+can then have an assignment operator, operator=, that (1) accepts a right-hand
+side value (10, in our example), applies D's conversion from int to std::string,
+and finally places the std::string into the internal metadata storage.
+
+Again, though, why couldn't we have done that with the call operator? Well, we
+could have, but with some unpleasant surprises. With our existing call operator,
+you can write either of the following and get the expected result:
+
+   int i = node(A,B,C,D);
+   auto j = node(A,B,C,D); // j is an int :-)
+
+The right-hand side produces an int. Not something that's merely convertible to
+an int, but a hard int. So, "auto" in the second line evaluates as you'd expect.
+
+Consider, however the bracket form:
+
+   int k = node(A,B,C)[D]; // ok, RHS is in fact convertible to int
+   auto L = node(A,B,C)[D]; // um, well... :-/
+
+Surprise! The second expression may compile, but "auto" doesn't give an int -
+as a user would probably anticipate it would. Instead, it gives an unexpected
+type: that of the smart object we spoke of. The object that stores information
+that we need in order for it to behave as an lvalue.
+
+The smart object produced by operator[] is equipped with assignment from, and
+conversion to, both the raw metadatum type std::string, and the type (here, int)
+that's encoded in the meta_t object. (Recall that D in our examples here is a
+meta_t<int>.) So, one *could* write this:
+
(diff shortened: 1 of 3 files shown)


staleyLANL:
OK, I'm not sure what's going on here. It says 88 files changed. I added one and changed about 3. I may have made some sort of git mistake.

staleyLANL:
Maybe the other PR that was approved earlier today didn't contain all the material I thought it did. I'd had two PRs in, and removed what I thought was the first, outdated one (because the second should have included the first's material). Anyway, the main new stuff is supposed to be a couple new classes near the bottom of node detail, and a new file node subscript.

whaeck:
@staleyLANL I think the previous one did not get pulled into master but in feature/diagnostics-and-more.

staleyLANL:
@whaeck Ah, so somehow the PR I merged earlier was slated to go into the other branch, not master. Then, since the current one was an extension of that - and apparently I did say to put it into master - then master thinks it's getting both the brand-new stuff and the stuff that only made it into diagnostics-and-more earlier? My main concern is that my work doesn't vanish somehow. :-) The new material should basically be the bottom part of detail.hpp, and everything in subscript.hpp, both in Node. See subscript.hpp's long comment for an explanation of the what/why/how. I think you can use this stuff right away.

whaeck:
@staleyLANL To solve this, you may want to pull feature/diagnostics-and-more into master first, and then update this one to the new master once that is done.

src/GNDStk/Node/src/subscript.hpp
```diff
@@ -0,0 +1,275 @@
...
+   n(axis)
+
+returns a container (default std::vector) of nodes. We could have child_ref in
+this case contain, for example, a std::vector<std::reference_wrapper<node>>,
+or even just a std::vector<node*>. As an lvalue, though, what would assignment
+mean? Replace each container element, requiring that the assignment's RHS has
+the same container size? Remove all the existing axis nodes of the parent node,
+completely replacing them with the nodes in the assignment's RHS? Until/unless
+we decide what's best, I'll have operator[] for allow::many child_t objects do
+what operator() does. So, fow now: defined, but not usable as an lvalue.
```
whaeck, line 177: Now, if I wrote the following:
```
n[ axis, [] ( const auto& node ) { return node( index ) == "0"; } ]
```
would that return an lvalue so that I can actually modify that node?

whaeck: I'm inclined to agree with you that if the operator[] returns more than one node, we should not treat the container as an lvalue. Having the vector contain reference_wrappers is a more sensible solution - unless we establish the rule that ONLY a single node can be an lvalue.

whaeck (approved):
Now that was a lot easier to review :-)

Do you have some tests associated to this? You may want to add it to the todo list for later, or add on a test file if you have time.

Other than that, I just have a question on using the operator[] with a filter to retrieve a single instance of a many node.

staleyLANL:
> Now that was a lot easier to review :-)
> 
> Do you have some tests associated to this? You may want to add it to the todo list for later, or add on a test file if you have time.
> 
> Other than that, I just have a question on using the operator[] with a filter to retrieve a single instance of a many node.

I wrote some informal tests, but haven't integrated any tests into the formal test quite yet. That's of course on my to-do list; my intention had been to just get this in quickly so that it can be used.

src/GNDStk/Node/src/subscript.hpp
staleyLANL: I'm thinking more and more than the call operator should also return a vector of reference wrappers, not a vector of copies of the nodes.

And it goes without saying (although, well, I'm saying it) that that applies to the case where you use child_t<void>, not child_t<some type other than void>. In the latter case, conversion from node to some user type must be provided, and the conversion is applied to create each element of the output vector. So stuff is created on-the-fly, and the reference idea is not applicable.

staleyLANL: > Now, if I wrote the following:
> 
> ```
> n[ axis, [] ( const auto& node ) { return node( index ) == "0"; } ]
> ```
> 
> would that return an lvalue so that I can actually modify that node?

I think you want to *downgrade* axis to an allow::one, and *add* the lambda:
```
n[ --axis + lambda ];
```
Let me know if this indeed works. While I think I got `operator[]` right, I don't have full tests for it yet.

staleyLANL: Of course you'd presumably actually assign that to something:
```
n[ --axis + lambda ] = foobar;
```
Note that my `operator[]` is tricked out so that it can assign to either a value of the *type* given to (in this case) `axis`, or to anything that's implicitly convertible to that type (like, `int` to `double`) - or even to a string, if for whatever reason you'd like to put in the raw string directly instead of using the type you've associated with the keyword.

whaeck: Ah, I forgot that I needed a '+' to add the filter. I'll give it a try when I update the v1.9 branch.

staleyLANL: Re: my above comment edit. If I ever make a comment that sounds stupid, let me know, and I can probably fix it. It's probably markdown's fault!!. Unless it's my fault, in which case I'll still try to fix it.

whaeck: So if I write:
```
auto filter = [] ( const auto& node ) { return node( index ) == "0"; };
n[ Axis{} / --axis + filter ] = Axis( 0, "someLabel", "someUnit" );
```
I can use one of the higher level components to change the node itself (if I have a convert function handy, that is)?

whaeck:
@staleyLANL I'll go and merge this tomorrow once I've done a few tests in the v1.9 branch.

src/GNDStk/Node/src/subscript.hpp
staleyLANL: By the way, regarding my `--` and `++` operators on keywords: Strangely enough, they *don't* actually modify the object itself, like how `--i` decrements to actual contents of the integer `i`. My keyword arithmetic always produces brand-new objects. (In many cases it couldn't *not* do that, because the arithmetic changes a template type!). So, `--child_t` has no effect on the original, and the original can even be `const`. Note that the language does indeed allow this behavior - of increment and decrement not changing the object to which they're applied. Of course I wouldn't do it if the language didn't allow it.

staleyLANL: I need to check if the remarks in my `foobar` example above actually work as I indent, re: the right-hand side being `std::string`. I may need to explicitly support the `char*` case, to prevent a general template function overload (from which an exact match can be produced) from being called, in place of the expected implicit `char *` to `std::string` conversion. Ah, technical stuff.

staleyLANL: To prevent any possible confusion... Be aware that a single minus (unary minus, `-`) in front of a keyword downgrades the *first* template argument - `type` - to `void`, meaning raw tree form rather than user-defined type. That applies to both `meta_t` and `child_t`. *Two* minus signs (predecrement, `--`) in front of a keyword downgrades the *second* template argument of a `child_t` - making it `allow::one`, regardless of its previous value. (The predecrement doesn't apply to `meta_t` objects, which don't have the `one` versus `many` distinction.)

There's indeed a pattern to all this funny stuff I wrote. A user just needs to stare at it for a bit, in order for it all to make sense. If it still doesn't make sense after some study, then let's talk about it. :-)

staleyLANL: > So if I write:
> 
> ```
> auto filter = [] ( const auto& node ) { return node( index ) == "0"; };
> n[ Axis{} / --axis + filter ] = Axis( 0, "someLabel", "someUnit" );
> ```
> 
> I can use one of the higher level components to change the node itself (if I have a convert function handy, that is)?

Offhand that looks correct, and it *should* work. Let me know if it doesn't! The operator business is fairly recent, and the bracket operator is brand-new, so there could always be a glitch here or there.

Of course we can, as I illustrated in my sample high-level Axis, associate an existing node with a high-level structure, and have a `save()` or other operation that changes the node automatically. We'll probably want to allow more than one way to do things.

Your comment reminds me: we should have a discussion sometime of what I did provide in the way of built-in `convert` functions. Roughly speaking: a handful of basic ones are provided for metadata (string/type conversions), and none are provided for children (node/type) conversions. None, because I can't imagine any meaningful default for that case; it depends on what's in somebody's data structure.

staleyLANL: It looks like all this operator business has proven to be even more useful than I anticipated it might be when I wrote it. That's a good thing.

staleyLANL:
@whaeck @jlconlin Looks like this has one approval, at least, but I still don't have merge permissions. What do you think about going ahead and merging? I'll have additional material soon, and probably more tests for the above, by week's end, at which time I can put in another PR.

PR #16, opened 2020-11-19
Feature/diagnostics and more

staleyLANL:
Want to merge this into master.


whaeck (approved):
I looked over all files and recognized most of the changes from before. The other changes are cosmetic changes (e.g changes to catch (...), etc.)

I think we're good here. At least we'll be able to work with a clean slate once this is in master.

staleyLANL:
@whaeck I see you approved, but it says merging is blocked. Not sure why; I was able to do the other merge (that apparently wasn't into master).

PR #17, opened 2020-12-04
Feature/misc functionality

staleyLANL:


staleyLANL:
@jlconlin @whaeck If you get a chance to look at this soon, that would be great. I have some new material that should be ready in a day or two, but probably best that I do that as another PR, not added to this one.

src/GNDStk/Node/src/call-child.hpp
```diff
@@ -24,15 +24,39 @@ decltype(auto) operator()(
...
    try {
-      // As above, don't need or want:
-      //    if (kwd.name == "")
-      //       detail::apply_converter<TYPE>{}(kwd,*this);
-      return child(kwd.one(), detail::label_is(label), found);
+      // total filter
+      auto filter = [kwd,label](const Node &n)
+         { return kwd.filter(n) && detail::label_is(label)(n); };
+      // --kwd: child_t<...,allow::one,...>
+      return child(--kwd+filter, found);
```
whaeck, line 33: Isn't it nice when things come together :-)

src/GNDStk/Node/src/call.hpp
```diff
@@ -41,17 +52,20 @@ decltype(auto) operator()(
...
+      // ""?
       if (kwd.name == "")
          detail::apply_converter<TYPE>{}(kwd,*this);
-      return child(-kwd.one(), detail::label_is(label), found)
-                  (std::forward<KEYWORDS>(keywords)...);
+      // total filter
+      auto filter = [kwd,label](const Node &n)
+         { return kwd.filter(n) && detail::label_is(label)(n); };
+      // -(--kwd): child_t<void,allow::one,...>
+      return child(-(--kwd)+filter, found)(std::forward<KEYWORDS>(keywords)...);
```
whaeck, line 62: OK, so I know that --kwd degrades the many state of the keyword to one, but what does the negation of the result do?

src/GNDStk/Node/src/compare.hpp
```diff
@@ -0,0 +1,80 @@
...
+   if (one.metadata.size() != two.metadata.size()) return false;
+   if (one.children.size() != two.children.size()) return false;
+
+   // metadata
+   if (std::set<metaPair>(one.metadata.begin(), one.metadata.end()) !=
+       std::set<metaPair>(two.metadata.begin(), two.metadata.end()))
+      return false;
+
+   // ------------------------
+   // OK, not fast/easy :-(
```
whaeck, line 38: Nah, it'll work ;-)

I don't think users will be using this feature anyway. It's more for testing purposes.

src/GNDStk/Node/src/count.hpp
```diff
@@ -0,0 +1,135 @@
...
+// -----------------------------------------------------------------------------
+// Node.count
+// -----------------------------------------------------------------------------
+
+// ------------------------
+// string[,filter]
+// ------------------------
+
+template<class FILTER = detail::noFilter>
+size_t count(
```
whaeck, line 61: Nice, I'll use this when verifying the number of child nodes.

src/GNDStk/Node/src/special.hpp
```diff
@@ -150,7 +150,7 @@ const std::string &comment(
 ) const {
    size_t count = 0;
 
-   for (auto c : children)
+   for (auto &c : children)
```
whaeck, line 153: Been there, done that

src/GNDStk/Node/test/subscript.test.cpp
```diff
@@ -0,0 +1,27 @@
+
+#include "catch.hpp"
+#include "GNDStk.hpp"
+using namespace njoy::GNDStk;
+
+SCENARIO("Testing GNDStk Node operator[]") {
```
whaeck, line 6: I assume we're finishing this later?

src/GNDStk/basic/src/child.hpp
```diff
@@ -116,7 +116,7 @@ inline const child_t<void,allow::many>
    axis                            ("axis"                            );
 inline const child_t<void,allow::one>
    background                      ("background"                      );
-inline const child_t<void,allow::one>
+inline const child_t<void,allow::one> // <== fixme should this be a many?
```
whaeck, line 119: Good question, we'll have to check in the specs

src/GNDStk/utility.hpp
```diff
@@ -125,6 +126,13 @@ inline std::string diagnostic(
 // Shortcuts to njoy::Log functions
 // -----------------------------------------------------------------------------
 
+// Convenient, namespace-scope booleans for fine-tuning diagnostic output
+inline bool notes     = true; // general info messages
+inline bool debugging = true; // debug calls
+inline bool warnings  = true; // warnings
+inline bool context   = true; // any of our context-printing info messages
```
whaeck, line 133: Nice, I can now switch the warnings off for testing when I write small GNDS snippets

whaeck (approved):
Good for me. Just a few questions but nothing that shouldn't stop us from merging.

src/GNDStk/Node/src/call-child.hpp
staleyLANL: Definitely. Build a solid foundation, and adding new features becomes easier and easier.

src/GNDStk/Node/src/call.hpp
staleyLANL: There's a system to these operators. :-)

ONE minus (unary minus operator) downgrades the FIRST template argument.
TWO minuses (predecrement operator) downgrades the SECOND template argument.

So, one minus takes a `meta_t<type,...>` or `child_t<type,...>` and downgrades to `meta_t<void>` or `child_t<void>`.
Which, as you know, means to retrieve the `meta_t` or `child_t` in their raw tree form: const or non-const string or node reference.

Here's what you're really seeing in the code block you referenced. We're traversing a query sequence, say `node(A,B,C,D,E)`. The code does this by "peeling off" the first argument in order to get to the child node (or the current `*this` node) represented by the first argument. It downgrades to `<void>`, even if the `child_t` has a type, because we need to stay in tree form (so we can continue going to child nodes) until we get to the very last argument - E, in my example. In this way, non-`<void>` `child_t` objects can appear in sequence, without the user having to explicitly downgrade them just because they're not at the end.

As for the `--` in the referenced code, it's because - in that variation of `operator()` - we have something like `node(...,child_t,"foobar",...)`, for which GNDStk's interpretation is to look for the specific case of the child node that has metadatum `"label" == "foobar"`. For that interpretation, GNDStk allows even an `allow::many` `child_t` (which normally wouldn't be allowed before the end of the sequence), because it assumes that the extra label-match condition means there's probably just one, not many.

Does that make sense?

staleyLANL: A bit of history: the unary minus is something I put in a long time ago, well before all the other operators, because I needed to do just what I described: stay in `<void>` form - raw tree form - until we get to the end.

The original thinking was: I'm calling this my "smart keyword" system, because it can deal with custom user types.

A "dumb keyword", on the other hand, would just have the `<void>` meaning - stay in tree form, without the nice type system.

So, unary MINUS takes a smart keyword and "SUBTRACTS from its intelligence", making it a dumb keyword. :-p

src/GNDStk/Node/src/compare.hpp
staleyLANL: Yes, that's the thought, and is why I decided not to spend too much time trying to optimize that particular capability.

src/GNDStk/Node/src/special.hpp
staleyLANL: And in this case it wasn't just an issue of inefficiency from copying. The code (if instantiated!) didn't compile, because those things are unique pointers - which don't copy!

src/GNDStk/Node/src/count.hpp
staleyLANL: Good to hear. Your needs, with regards to the high-level interface, are driving much of the new things I'm writing.

src/GNDStk/Node/test/subscript.test.cpp
staleyLANL: Correct.

src/GNDStk/basic/src/child.hpp
staleyLANL: There are probably a handful of these things that are wrong.

src/GNDStk/utility.hpp
staleyLANL: I thought about writing "debugs" and "contexts" in order to have consistent language, although "debugs" doesn't come across as being a legitimate plural *noun*. What do you think?

PR #18, opened 2020-12-15
Feature/multi query

staleyLANL:
Here's a ton of new material that I want to get in sooner rather than later, before this PR would otherwise become too big. There's much more I want to do, and much of the new work needs formal tests. All that's on my to-do list, and will go into another PR at some point. For now, much of this should help @whaeck in particular with some of the material he's been writing for the high-level interface.


src/GNDStk.hpp
```diff
@@ -81,5 +82,36 @@ namespace GNDStk {
...
+nodes in their original tree form. Note that basic:: itself brings in meta_t
+and child_t objects from its own nested meta:: and child:: namespaces. Those
+are separate because there's a small amount of overlap between allowable GNDS
+metadatum names and node names. With this scheme, if a name you wish to use
+isn't one of the overlapping names, just use it. If it is, then prefix with
+meta:: or child:: depending on which one you want.
+*/
+
+namespace core {
+   using namespace GNDStk;
```
whaeck, line 112: We will (soon) have the standardised versioned namespaces in the GNDStk namespace as well, so we may want to move the code that is currently in the GNDStk namespace to the core namespace and it's own core.hpp file.

src/GNDStk/utility.hpp
```diff
@@ -67,6 +67,14 @@ class noFilter {
    bool operator()(const NODE &) const { return true; }
 };
 
+// failback
+inline void failback(std::istream &is, const std::streampos pos)
+{
+   is.clear();
+   is.seekg(pos);
+   is.setstate(std::ios::failbit);
+}
```
whaeck, line 76: One wonders where you got this from ;-)

src/GNDStk/Node/src/call-keywords.hpp
```diff
@@ -0,0 +1,159 @@
...
+      // Process tup's <0>
+      // We *don't* simply write:
+      //    std::make_tuple(operator()(std::get<0>(tup),head_found)),
+      // because doing so would decay references to non-references. That would
+      // mean, ultimately, that meta_t<void> and child_t<void> instances inside
+      // our keywords<...> objects, which are returned by our | (or) operators,
+      // would no longer produce [possibly const] *references* to std::string
+      // and Node, respectively, but would surreptitiously lose the references.
+      // We don't want to lose the references, though, because our multi-query
+      // system is far more awesome if we keep them.
```
whaeck, line 36: I agree, it's awesome ;-)

I can't wait to use it once the std::optional support is present as well.

whaeck (approved):
The "typed" namespace has been removed and the core namespace has been introduced (as I noted below/above, we may want to move this to its own core.hpp file at some point).

Addition of the multi-query system (to quote the Lego movie: everything is awesome!), some new SFINAE to help with the introduction of std::optional and some cleanup.

This is good to go.

PR #19, opened 2020-12-17
Made multi-query actually work. :-) Some of that was tricky.

staleyLANL:
Made its error handling consistent with that of other query functions.
Gave it good, informative diagnostics too.
Adjusted operator() function signatures so new overloads all work properly.
Added some helper functions for certain diagnostic printing.
size_t is really std::size_t.
Needed char[N] in one of the traits classes.
Small comments and code tweaks here and there.

### DIFF
--- a/src/GNDStk.hpp
+++ b/src/GNDStk.hpp
@@ -98,14 +98,13 @@ gives the same effect as writing this:
    using namespace njoy::GNDStk::basic;
 
 Meaning: We consider our "core interface" to consist of everything in GNDStk::
-proper, plus our basic:: set of meta_t and child_t keywords. Those are the
-meta_t and child_t objects with <void> type, so that they return metadata and
-nodes in their original tree form. Note that basic:: itself brings in meta_t
-and child_t objects from its own nested meta:: and child:: namespaces. Those
-are separate because there's a small amount of overlap between allowable GNDS
-metadatum names and node names. With this scheme, if a name you wish to use
-isn't one of the overlapping names, just use it. If it is, then prefix with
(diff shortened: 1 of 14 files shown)


staleyLANL:
Complete, working example:

```
#include "GNDStk.hpp"
using namespace njoy::GNDStk::core;

int main()
{
   tree t("n-094_Pu_239.xml");

   auto info = t(
      // as usual, --s downgrade allow::many to allow::one for query
      reactionSuite,
      PoPs,
      chemicalElements,
      --chemicalElement,
      isotopes,
      --isotope + [](const auto &node){ return node(symbol) == "Pu236"; },
      (
         symbol | // meta
         A |      // meta
         nuclides // child
      )
   );

   assert(std::get<0>(info) == "Pu236");
   assert(std::get<1>(info) == "236");

   // <nuclides> has no metadata
   assert(std::get<2>(info).metadata.size() == 0);

   // <nuclides> has one <nuclide> child
   assert(std::get<2>(info).children.size() == 1);
   assert(std::get<2>(info).children[0]->name == "nuclide");
}
```

whaeck:
@staleyLANL You should make that into a test ;-)

whaeck:
@staleyLANL I tried the multi-query objects with the v1.9 branch, it works as advertised. I did realise that I'll have to test the node references in the case where I have an interface with a node that has children (like axes and constant1d).

staleyLANL:
@whaeck By the way, with the multi-query object being built with bitwise-or rather than comma, you should no longer need the parentheses.

whaeck (approved):

PR #20, opened 2020-12-22
std::optional handled in queries.

staleyLANL:
Some improvements and increased clarity re: diagnostic fine-tuning flags.
Miscellaneous small tweaks to formatting and comments.
Slightly simplified/shortened some SFINAE and other constructs.
Improvements to diagnostic printing; less is hardcoded.

### DIFF
--- a/src/GNDStk/Node/src/call-keywords.hpp
+++ b/src/GNDStk/Node/src/call-keywords.hpp
@@ -33,9 +33,7 @@ template<
    class TYPE, allow ALLOW, class CONVERTER, class FILTER,
    class SECOND,
    class... TAIL,
-   class = typename std::enable_if<
-      detail::is_string_or_regex<SECOND>::value
-   >::type
+   class = typename detail::is_string_or_regex<SECOND>::type
 >
 auto operator()(
    const std::tuple<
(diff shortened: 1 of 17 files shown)


whaeck:
I updated the v1.9 branch and it now uses std::optional in the multi-query objects, it compiles and tests continue to pass. I'm a happy man :-)

src/GNDStk/Node/src/call-keywords.hpp
```diff
@@ -33,9 +33,7 @@ template<
    class TYPE, allow ALLOW, class CONVERTER, class FILTER,
    class SECOND,
    class... TAIL,
-   class = typename std::enable_if<
-      detail::is_string_or_regex<SECOND>::value
-   >::type
```
whaeck, line 38: My SFINAE-fu is weak. Taking away the enable_if does not change anything? Just a question to understand better.

whaeck (approved):
Diagnostic messages are updated, some SFINAE changes (removing enable_if) and adding the std::optional support (the main topic of this pull request).

The code has been applied in feature/v1.9 and it works well. Good to go - just one question to improve my SFINAE-fu.

src/GNDStk/Node/src/call-keywords.hpp
staleyLANL: It doesn't change anything in this case at least. With SFINAE, you just need to trigger some sort of SF (substitution failure), generally in the return type of a function template. So, something that, roughly speaking, "doesn't work", or "doesn't make sense", when something specific is tried for a template parameter.

`enable_if` is convenient because it takes a `bool` as the first argument, which means you can do things like `&&` and `||` when you need them. If `true`, something is there (in class `enable_if`) for `type`. If `false`, then `type` isn't there at all - triggering the substitution failure.

In (some of) my SFINAE-supporting classes, I happen to have a `type` already, for other purposes. So, for those, I can just go the direct route instead of involving `enable_if`. My `is_string_or_regex` has a `type` if, and only if, the template parameter is `std::string` or a moral equivalent (say, `char*`), or a `std::regex`. It turned out that I needed that in order to handle those string-like things such as `char*` properly.

Of course when I speak of `type`, it doesn't need to be called `type`. The key idea is that we're pulling `::something` out of a class template for some template substitution, and SFINAE is all about whether the `::something` works or not.

PR #21, opened 2020-12-24
Much more capable and complete operator[].

staleyLANL:
This should help immensely with the high-level interface.
Full tests for operator[].
That was tons of careful work. I need a vacation.
Fortunately, it's Christmas Eve right now.

### DIFF
--- a/src/GNDStk/Node/test/subscript.test.cpp
+++ b/src/GNDStk/Node/test/subscript.test.cpp
@@ -1,26 +1,691 @@
 
 #include "catch.hpp"
 #include "GNDStk.hpp"
-using namespace njoy::GNDStk;
+using namespace njoy::GNDStk::core;
+
+
+// -----------------------------------------------------------------------------
+// Custom types
+// -----------------------------------------------------------------------------
+
+// Really? major and minor are #defined in some system header??
+#undef major
+#undef minor
+
+
+// ------------------------
+// version_t
+// used with meta_t
+// ------------------------
+
+struct version_t {
+   int major = -1;
+   int minor = -1;
+   int patch = -1;
+   version_t() { }
+   version_t(const int a, const int b, const int c) :
+      major(a), minor(b), patch(c)
+   { }
+};
+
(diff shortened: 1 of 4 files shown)


staleyLANL:
@jlconlin @whaeck I'll probably have more material upload in a day or two, so feel free to take a look at this whenever you get the chance.

staleyLANL:
@jlconlin @whaeck If and when this is good to go, I have more material I can upload.

src/GNDStk/Node/test/subscript.test.cpp
```diff
@@ -1,26 +1,691 @@
...
+            // Actually writes into the tree...
+            EvaluatedNode[myversion] = "123.456.789";
+            CHECK(tree(reactionSuite,styles,evaluated,basic::version) ==
+                  "123.456.789");
+            // can still use as rvalue, and implicitly converts to string...
+            std::string foo = EvaluatedNode[myversion];
+            CHECK(foo == "123.456.789");
+
+            // Also actually writes into the tree...
+            EvaluatedNode[myversion] = version_t(12,34,56);
```
whaeck, line 289: I assume the prerequisite for this to work is to have an ostream operator>> defined for version_t?

Builtin types like double, int, etc. have this by default. We should discuss precision in floating points at some point.

whaeck (approved):
Some renaming arguments, assigning default arguments. The main work here is operator[], tests were added for this to illustrate the functionality. Good to go for me.

src/GNDStk/Node/test/subscript.test.cpp
staleyLANL: Re: version_t, the prerequisite is that it have the usual GNDStk `convert()` business. However, my built-in `convert()` functions for `meta_t` eventually fall back on stream I/O if all else fails.

staleyLANL: Let's come up with a short example in which we write a couple of small sections of GNDStk data, and want to set precision individually in each one.

PR #22, opened 2021-01-08
Major changes: operator[] internals improvement; "values" completely …

staleyLANL:
…redone.

Consolidated + shortened a considerable amount of material related to operator[].
Reformulated the child_t previously called "values".
"values" now just retrieves a regular node.
Inside values, use e.g. numeric<double> to get vector<double> of the numeric data.
In short: replace foo(values<>) with foo(--values,numeric<>).
Important: values is an allow::many; it does *sometimes* appear multiple times.
So, metadata in GNDS/XML <values> nodes is now available in the usual manner.
Shortened detail:: code related to values/pcdata/numeric as described above.
Changed the GNDSTK_MAKE_CHILD macro so that instances of its use are shorter.
As a consequence, shortened instances of its use. :-)
As usual, code and comment tweaks here and there.
Removed the convert_pcdata_text object; I don't think anyone needs direct access.
Fixed some fixmes.
Separated one/many in basic/src/child.hpp; this shortened the file considerably.
Re: the above, adjusted all relevant test codes.


staleyLANL:
@jlconlin @whaeck Any thoughts on this PR? I plan to have another one up by tomorrow evening, before the long weekend, so if you have a chance to take a look, that would be great. :-)

src/GNDStk/Node/src/detail.hpp
```diff
@@ -568,41 +507,88 @@ class child_ref<
...
+
    // size
    std::size_t size() const
    {
       return childNodePtr.size();
    }
 
    // operator[]
-   auto operator[](const std::size_t n) const
+   decltype(auto) operator[](const std::size_t n) const
```
whaeck, line 526: I got bitten by that one a couple of times myself :-)

whaeck (approved):
Mainly changes in comments, some reorganising in the children. Update to tests where required.

Good to go.

src/GNDStk/Node/src/detail.hpp
staleyLANL: What I previously had actually worked, so there wasn't an error. For this PR, though, I consolidated some things, and the consolidation caused it to be necessary to change `auto` to `decltype(auto)`. Note that the function in question can return either a reference or a non-reference, depending on the `if constexpr`. The reference possibility is what necessitated the `decltype`.

PR #23, opened 2021-01-14
New query feature. TYPE instance in query objects.

staleyLANL:
Query of the form `node(...,somechild,"somestring")` is more flexible.
...string of "" means "label isn't there, or is actually empty".
...useful in some GNDS files to distinguish `<values>` nodes under one parent.
Classes `meta_t<TYPE>` and `child_t<TYPE>` now contain an object of type TYPE.
Among other things, this will enable an "optional with default" implementation.
...That was the main motivation for this work.
Substantial rework of constructors for `child_t`.
...It allows a TYPE object; but otherwise is more restricted.
...Use our existing operators to regain flexibility; that's easier anyway.
"Keyword builders" keyword.meta()/keyword.child() cleaner, more flexible.
Various changed or brand-new tests, re: `meta_t` and `child_t` modifications.
Operators `~` and `!` to enable/disable "can be top-level node" indicator.
Functions `top()` and `top(bool)` for `child_t`
The "top-level node" idea still needs some work; more about this another time.
As usual, slight tweaks to comments, formatting, etc. here and there.


staleyLANL:
@whaeck I went ahead and added these modest one/many changes to the existing PR, as I'm working on a completely different thing this week (a built-in "optional, with default" capability), and didn't want to clutter that with the one/many tweaks.

src/GNDStk/Node/src/add.hpp
```diff
@@ -38,7 +44,7 @@ bool add(
    std::string to;
    try {
       if (opt)
-         converter(opt.value(),to);
+         converter(*opt,to);
```
whaeck, line 47: I would have kept the use of value() since it is more explicit than just using operator*. value() also throws an exception if used on an empty optional. Granted, since we check before using operato*, we're sure we won't get in trouble.

src/GNDStk/child_t/src/operator.hpp
```diff
@@ -16,6 +16,42 @@ inline auto operator-(const child_t<TYPE,ALLOW,CONVERTER,FILTER> &kwd)
...
+//
+// We don't expect much use of these, but someone may occasionally find the
+// first (more likely) or the second (less likely) to be helpful. The unary
+// operators that we chose for these purposes seemed like the best, given a
+// limited selection. One could think: (T)ilde for (T)op; Not for Not.
+// -----------------------------------------------------------------------------
+
+// operator~
+template<class TYPE, allow ALLOW, class CONVERTER, class FILTER>
+inline auto operator~(const child_t<TYPE,ALLOW,CONVERTER,FILTER> &kwd)
```
whaeck, line 31: While I like operator overloading, "assigning" a different meaning can be dangerous. Again, good documentation is required. I was wondering here if this couldn't give issues with destructors?

Why not use an actual function name instead of an operator overload?

whaeck (approved):
I'm just a bit concerned about going overboard on operator overloading, especially without some proper documentation. Call me old fashioned ;-)

jlconlin:
> I'm just a bit concerned about going overboard on operator overloading, especially without some proper documentation. Call me old fashioned ;-)

Yeah, me too.

src/GNDStk/child_t/src/operator.hpp
staleyLANL: I'll make a note to consider this for the next change. The "top-level" business needs some work, anyway, and isn't likely to be something that someone would need to change. So, perhaps I will just make these into regular functions. I didn't actually even think about the destructor issue when I was writing this, but I think the difference is that tilde as the destructor is a member function, while this one isn't, so they're unambiguous.

staleyLANL:
> I'm just a bit concerned about going overboard on operator overloading, especially without some proper documentation. Call me old fashioned ;-)



> > I'm just a bit concerned about going overboard on operator overloading, especially without some proper documentation. Call me old fashioned ;-)
> 
> Yeah, me too.

I'll second that. Wait, third that. I'd planned to begin documentation several weeks ago, but @whaeck 's prototyping of some high-level functionality got me thinking about changes that would help immensely with the high-level interface, and I've wanted to get those out of the way. At the moment, I'm working on an "optional with default," which I anticipate will be done by (or before) week's end, and then have one other thing I'd like to do for the benefit of the high-level interface. I'll make a commitment to begin writing documentation in earnest no later than the second week of February. Maybe the first, if the current work goes reasonably fast.

src/GNDStk/Node/src/add.hpp
staleyLANL: Sure, I'll go ahead and change those back to `value()`. As you said, I check anyway, so `operator*`'s potential for undefined behavior won't ever be realized. However, as you said, value() is more explicit. I was trying to be more compact, but I'd agree that `value()` really is clearer.

PR #24, opened 2021-01-21
"Optional, with default" class.

staleyLANL:
New class: defaulted, to represent an "optional, but with a default" concept.
Should be of great value for the high-level interface.
Important note: not yet given special treatment elsewhere in GNDStk.
The above is next on the agenda.
Complete tests for the new class.
Re older work: for std::optional, changed `*` calls to `value()` calls.


staleyLANL:
@jlconlin @whaeck I've been working on this pretty much nonstop. There's a bunch of new work and tests. @whaeck in particular, this deserves a careful look, so I can be sure it'll do what you need for the high-level interface. Note that I don't yet have special treatment of the new class by GNDStk, as I do with std::optional. That's next on the agenda. For now, then, it's just the class and all its functionality, as well as tests.

whaeck:
@staleyLANL OK, I'll have a look in detail as soon as I can. I did skim the code quickly. The defaulted class seems to mimic the std::optional interface, which I like.

staleyLANL:
> @staleyLANL OK, I'll have a look in detail as soon as I can. I did skim the code quickly. The defaulted class seems to mimic the std::optional interface, which I like.

@jlconlin @whaeck The new optional-with-default should be ready for prime time now. I've completely integrated it into both the queries and the `add()` functions. (But I do need to know if it behaves the way you're looking for, in the high-level interface.) Handling it in the `add()` functions raised some questions about the exact design of everything there, and I ended up doing lots of work on those. It's clear that good documentation is needed, and I'm planning to start writing documentation next Monday, even if there are still a few coding things I'd love to do. More code work will be interspersed with documentation work as the need arises.

staleyLANL:
@whaeck @jlconlin Any chance to look at this? I have some new things - not really belonging in this particular PR - that I'm hoping to upload before next Monday, as I'll be starting on the documentation then.

jlconlin:
A

> @whaeck @jlconlin Any chance to look at this? I have some new things - not really belonging in this particular PR - that I'm hoping to upload before next Monday, as I'll be starting on the documentation then.

Without documentation, I can't really tell what you are doing. [^1] I've been asking for documentation for months and haven't seen one word of it. I'm not really motivated to continue to look at these Pull Requests until I see some documentation to give me some idea of what is going on.


[^1] Of course I can figure it out.

jlconlin:
> A
> 
> > @whaeck @jlconlin Any chance to look at this? I have some new things - not really belonging in this particular PR - that I'm hoping to upload before next Monday, as I'll be starting on the documentation then.
> 
> Without documentation, I can't really tell what you are doing. [^1] I've been asking for documentation for months and haven't seen one word of it. I'm not really motivated to continue to look at these Pull Requests until I see some documentation to give me some idea of what is going on.

Sorry I got on my soapbox.

src/GNDStk/Node/src/add-child.hpp
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// child_t<optional>, optional
+// Returns bool: was something added?
+template<
+   class TYPE, allow ALLOW, class CONVERTER, class FILTER,
+   class T,
+   class = typename std::enable_if<std::is_constructible<TYPE,T>::value>::type
+>
+bool add(
```
jlconlin, line 164: I'm not a big fan of this signature. Sometimes the `add` function returns a `Node&` and sometimes it returns a `bool`. Someone (likely me) is going to be confused and upset that sometimes this does what he/she wants and sometimes it doesn't.

src/GNDStk/Node/test/add.test.cpp
```diff
@@ -79,7 +79,12 @@ SCENARIO("Testing GNDStk Node add()") {
       n.add(dbl);
       n.add(dbl,1.23);
       n.add(dim);
+      // these need the = TYPE on class T in the metadata node.add() functions:
       n.add(dim,{321,987});
+      auto optdim = keyword.meta<std::optional<dimensions2d>>("OptDim");
+      auto defdim = defaulted<dimensions2d>({1,2})/keyword.meta<>("DefDim");
+      n.add(optdim,{321,987});
+      n.add(defdim,{321,987});
```
jlconlin, line 87: I'm not sure what you are testing here. You `add` some things, but you don't verify that anything was added. In short, you are just verifying that something compiles. Am I misunderstanding what you are doing here?

src/GNDStk/defaulted.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+Note that you can give a defaulted<T> a hard value that happens to equal its
+default. If you do, then the value is considered to be a hard value - not the
+default. That is to say, the data structure does distinguish "has its default"
+versus "has a value that coincidentally happens to equal its default." It's an
+issue of how you construct and then modify (or don't modify) the object.
+*/
+
+template<class T>
+class defaulted {
```
jlconlin, line 24: This should be called `Defaulted` as it is a class.

src/GNDStk/defaulted/src/value.hpp
```diff
@@ -0,0 +1,69 @@
+
+// -----------------------------------------------------------------------------
+// Get the value; various forms
+// -----------------------------------------------------------------------------
+
+// ------------------------
+// Channeling std::optional
+// ------------------------
+
+// These are here largely so this class is more consistent with std::optional,
```
jlconlin, line 10: 👍🏻

src/GNDStk/defaulted/src/value.hpp
```diff
@@ -0,0 +1,69 @@
...
+//
+// Our current "optional, with default" class contrasts with std::optional<T>
+// in the following manner. We dispense with std::optional<T>'s conversion to
+// bool,and instead have a conversion to T that's always both unambiguous and
+// meaningful (a defaulted<T> always has a value in some sense - its default,
+// if nothing else). (As we'll see elsewhere, we'll also keep has_value() in
+// this class, with the meaning: "do we have a value other than the default?")
+
+// to T
+operator T() const
```
jlconlin, line 66: I'm not really sure what this function does, or how it would be called. 

```c++
defaulted< double > d{ 3.1 };

d.double(); // Is this what this operator would be?
```

src/GNDStk/defaulted/test/value.test.cpp
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      // ->
+      THEN("operator-> works as expected") {
+         CHECK(v->size() == 4);
+      }
+
+      // value()
+      THEN("conversion to T works as expected") {
+         CHECK(int(d) == 123);
```
jlconlin, line 34: Ah, here is where I see how the operator is used. It isn't clear from the function signature this is how it should be used. Perhaps that's my C++ ignorance showing.

jlconlin (changes requested):
If I understand correctly, you are trying to create an object that---when not given a value---has a default value prescribed. I can see where we would need to have such an object in GNDS. I worry that you are complicating things here. Either that, or I don't understand.

You have a default value and an optional value. Why not use a single data member instead of having both? If you have given the object a non-default value, why keep the default value around? I would think that the default value would be something known at compile time and not something that is read off of a data file somewhere. 

Please help me understand your reasoning behind all of this.

src/GNDStk/Node/src/add-child.hpp
```diff
@@ -0,0 +1,283 @@
+
```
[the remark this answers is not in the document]
jlconlin: Another issue with this is there are many different functions with every-so-slightly-different signatures. Do we really have to have different definitions of this? Again, this could be an issue with my understanding of the problem, but it feels unnecessarily complex.

src/GNDStk/defaulted.hpp
whaeck: I'm actually talking with Martin right now, and he was talking about making changes to the names to be more in line with our traditional style - at least with classes that are going to be publicly used. He wants to change that now, before he starts on the documentation next week.

staleyLANL:
> > A
> > > @whaeck @jlconlin Any chance to look at this? I have some new things - not really belonging in this particular PR - that I'm hoping to upload before next Monday, as I'll be starting on the documentation then.
> > 
> > 
> > Without documentation, I can't really tell what you are doing. [^1] I've been asking for documentation for months and haven't seen one word of it. I'm not really motivated to continue to look at these Pull Requests until I see some documentation to give me some idea of what is going on.
> 
> Sorry I got on my soapbox.

I'm putting the finishing touches on various global name changes and a few formatting/style issues, and will focus on the documentation beginning Monday. I think that everything - how it works, and my reasoning process behind it - will become much more clear after it's all documented.

src/GNDStk/defaulted.hpp
staleyLANL: And that, for two big reasons: (1) so I don't have to modify the documentation later, and (2) so users don't start writing one thing, based on the documentation, and then have the rug pulled out from under them. :-)

staleyLANL: `defaulted` is definitely changing to `Defaulted`, and a few other things too.

src/GNDStk/Node/test/add.test.cpp
staleyLANL: In this case, actually, I *am* just ensuring that it compiles. For a brace-enclosed initializer to automatically become a `std::vector`, in this context, requires some extra template chicanery. But I can certainly go in and also, as a secondary goal, make sure that the values were added.

src/GNDStk/Node/src/add-child.hpp
staleyLANL: This is actually something that I somewhat agonized over as well. Returning a Node reference tends to be the most useful; think, for example, of the so-called Builder Pattern. Where `std::optional` or `GNDStk::Defaulted` is involved, however, it isn't always guaranteed that something will be added - which we believe will be the correct behavior when and where we use those classes in the high-level interface.

An option would be to just use `bool` everywhere; then, `add()` variations that are guaranteed to add the thing, just always return `true`. Of course I could also always return `void`, but that's less useful. Always returning a Node reference, and throwing an exception if that can't be done, would probably not work well in the context of the use of optional and defaulted objects in the high-level interface. Yet another possibility would be a pointer (nullptr if nothing was added), or perhaps a key/value pair, one of them being bool, as some functions in the C++ standard library do with respect to container reads and writes.

Ultimately, I settled on something that's both maximally useful, and for which differences can be explained and justified (and be easy to remember) in the documentation. Given that I'll be starting the documentation next week, how about we see if I can explain it well enough that it makes sense? If not, then I think it's still early enough (read: no users yet) that it can be changed.

src/GNDStk/defaulted/src/value.hpp
staleyLANL: That just an implicit-conversion-to-T operator. The bottom line is that you'd be able to use a Defaulted<T> in, say, an equation, and it'll magically convert to a T if it needs to.

src/GNDStk/Node/src/add-child.hpp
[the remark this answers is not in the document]
staleyLANL: I can't really argue too much with this. One reason I split Node's `add()` functions into two files (for metadata adds, and for child-node adds) was to try to help, if only a little, with all of it making sense. The plethora of `add()` variations makes the capability very flexible and powerful; unfortunately, that came at the price of a non-trivial amount of SFINAE drain bramage in order to make various calls overload to the right place. (Not the wrong place, which would trigger other problems and result in the spectacular - in a bad way - compiler error messages that we all know and love.) I'm hoping that the flexibility and power of the `add()` functions will be worth it. And, again, the upcoming documentation should be and to help in this respect.

src/GNDStk/Node/src/add-meta.hpp
```diff
@@ -0,0 +1,243 @@
+
```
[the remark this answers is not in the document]
staleyLANL: Breaking down this and add-child into more files might possibly help in this respect. That can always be done in the future, without disturbing anything users are doing.

staleyLANL:
> If I understand correctly, you are trying to create an object that---when not given a value---has a default value prescribed. I can see where we would need to have such an object in GNDS. I worry that you are complicating things here. Either that, or I don't understand.
> 
> You have a default value and an optional value. Why not use a single data member instead of having both? If you have given the object a non-default value, why keep the default value around? I would think that the default value would be something known at compile time and not something that is read off of a data file somewhere.
> 
> Please help me understand your reasoning behind all of this.

I hope, and believe, that my implementation of `Defaulted` will prove to be the correct one. It would be absolutely fantastic if the default value could be a template parameter, but, unfortunately, limitations in the C++ language prevent that. A template parameter can be of integral type, we all know, but can't be of a floating-point type. (I'm not talking about `template<class T>` where `T` is going to be `double`. I'm talking about `template<double d>` - which isn't allowed. Or, more generally, templating on some custom class.)

In terms of keeping the default around, in addition to having the optional, here's a sketch of the reasoning. Consider, say, the following scenario, or something like it. Someone discovers that there's a systematic error of some of the data in some GNDS node. You want to (1) read the node into some particular data structure we've built for those nodes; (2) correct the errors in the data; and (3) write the node back out. Ideally, the newly-written, corrected node appears identical to the old one, except for the data corrections. We preserve everything else, lest somebody wonder why we changed something that wasn't, itself, related to the systematic error in some data.

Maybe some value (say, a metadatum in the main node) was "optional, with default." It may, or may not, be there. Read the main node, and that value (or not) will be in a Defaulted. With my implementation, the Defaulted "knows" if it was there or not. Correct other data in the main node, and write the main back out. If an explicit value was in the original, it'll get written as an explicit value in the corrected node. If it wasn't there, Defaulted knows this - and won't write it into the new one. So, something that didn't appear in the original node won't magically pop up (you know, with its default value!) in the new one. What if the data correction perhaps requires changing the value - originally not there, because it was its default - into something slightly different. That's fine too. Just assign to the new value, and Defaulted "knows" it was changed. So it wasn't there to begin with (meaning: had its default), was changed to something else (meaning: now has something else), and it'll get written. Which we'd want: it should be written in the corrected node, if it's no longer its default!

Does that makes sense? Basically, Defaulted is designed to be kinda sorta magic. It sees what you're doing with the value, and can treat it as you'd probably want in terms of I/O.

src/GNDStk/defaulted/test/value.test.cpp
staleyLANL: You probably just haven't written too many C++ conversion operators. Google, say, "c++ user defined conversion operator" to get as much information as you want. It's one of those things that any given person might or might not have used much in their own codes.

staleyLANL:
I've done some of the suggested changes, i.e. the defaulted to Defaulted, either here or in some additional code I have waiting for another PR. (Sometimes I lose track of what exact branch I'm putting things into, as I try to reduce complexity, and merge issues, by just adding changes to new work that I might already have been doing after putting in a PR.) Some of the concerns are quite valid, like the large number of add() functions and the need to document how+why. I *am* working on documentation, and also, on what I think will be some really high-value ideas for the high-level interface. (Given that @whaeck has started writing high-level classes, I wanted to try out some ideas, in that respect, before waiting too long.) Would approving this PR be fine at this point? Then I can quickly upload more, without worrying about too many branches floating around.

whaeck (approved):

staleyLANL:
@jlconlin @whaeck It looks like I'm unable at this point to merge this pull request. I think most questions have been addressed, and I've done some of the name changes and that sort of thing. Some might be in some new work, not this branch. The main outstanding issue is probably the question of there being so many `add()` functions, which I think is something that will work itself out. Documentation has started (along with other work); when I get to the adds, I think they can either be explained well, or if not, then that part of the code can get another look. Initial users would probably be doing data queries rather than data creation. What would you say about getting this in and moving forward to new work?

PR #27, opened 2021-03-01
Feature/python

whaeck:
Setting up CMake for using pybind11, also provides some rudimentary bindings for the Node operator and python unit tests as well.

To compile the python module, you must use the GNDStk.python in the make command explicitly (the python module is excluded from the all target to prevent it from compiling when you compile projects that use the C++ GNDStk).
For example:
make GNDStk.python -j8

To run the python unit test, you should go into the python subdirectory and just run the runtests.sh script.

### DIFF
--- a/python/src/core/Node.python.cpp
+++ b/python/src/core/Node.python.cpp
@@ -14,6 +14,7 @@ namespace core {
 
     // type aliases
     using Component = njoy::GNDStk::core::Node;
+    using ConstRefComponent = std::reference_wrapper< const Component >;
 
     // create the core component
     python::class_< Component > component(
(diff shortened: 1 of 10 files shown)


python/src/core/Node.python.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+// namespace aliases
+namespace python = pybind11;
+
+namespace core {
+
+  void wrapNode( python::module& module ) {
+
+    // type aliases
+    using Component = njoy::GNDStk::core::Node;
```
staleyLANL, line 16: It probably doesn't matter here, but Node is technically in GNDStk:: itself (for which core:: has a using).

staleyLANL:
Looks fine, as far as I can tell, though I'm not a Python expert.

whaeck:
The main goal is to set up the CMakeLists.txt for python bindings. The preliminary Node wrapper is there to test if everything compiles and runs in github actions (which is does).

Anyway, you'll have to approve this because I can't - I'm the author ;-)

staleyLANL:
@whaeck I can't for the life of me find an "Approve" button anywhere. :-/

whaeck:
There should be a "review" or "review changes" button on the top right. If you click on that, you get a comment field and the possibility to approve, request changes, etc.

staleyLANL (approved):
Approved.

whaeck:
I see you found the button ;-)

PR #29, opened 2021-03-25
Feature/misc high level

staleyLANL:
See various commit comments for this branch.
I think it's time for this stuff to get a review.
Note that json2class.cpp isn't part of GNDStk proper, but is just a utility code.
With luck it's reasonably clear, but it isn't meant for the masses.
I mean masses of people, not "masses" in a nuclear data sense.


whaeck:
@staleyLANL I've added an executable target in the CMakeLists.txt file. I've updated the branch to the latest master to resolve a conflict with the CMakeLists.txt file in the master branch. To build the autogen code, you can now use
```
make json2class.exe
```

Now, the code compiles and executes properly but the hpp files are empty. They were actually empty to begin with.

staleyLANL (commented):
Looks good!

staleyLANL:
@whaeck Is this good to go? I'd like to get it wrapped up before submitting a new PR.

whaeck (approved):
Since the assert is fixed in the next PR, this is OK.

PR #30, opened 2021-04-06
Gave class Node a parent-node pointer.

staleyLANL:
The pointer itself is private; use parent() to get a reference.
The code has some miscellaneous validity check regarding the parent pointer.
Those might be removed at some point; but they probably don't affect speed much.
Small change to json2class.cpp, re: assert() call.
Node's debugging output can optionally print Node's address and parent address.
Removed Tree::nodeType; was left over from de-templatization of Node and Tree.
Minor code tweaks here and there.
More-descriptive names for some detail functions.

### DIFF
--- a/src/GNDStk.hpp
+++ b/src/GNDStk.hpp
@@ -17,6 +17,7 @@
 
 // miscellaneous
 #include <algorithm>
+#undef NDEBUG
 #include <cassert>
 #include <cctype>
 #include <chrono>
--- a/autogen/json2class.cpp
+++ b/autogen/json2class.cpp
@@ -1026,7 +1026,9 @@ void make_class(
    // save content
    if (debugging)
       std::cout << "ndep.name == " << ndep.name << std::endl;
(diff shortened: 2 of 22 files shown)


src/GNDStk.hpp
```diff
@@ -17,6 +17,7 @@
 
 // miscellaneous
 #include <algorithm>
+#undef NDEBUG
```
whaeck, line 20: No. Don't do this.

If you undefine it, it will propagate everywhere and could/will have unintended consequences. The Eigen library for instance is a header only library that uses a lot asserts to check calculations in Debug mode. As a result, it is a lot faster in Release mode where -DNDEBUG is used as a compiler flag.

whaeck (changes requested):
Other than the fact that you should not be undefining NDEBUG, this looks good to me.

src/GNDStk.hpp
staleyLANL: OK, I changed it.

whaeck (approved):

PR #39, opened 2021-06-07
Resolving an ambiguous name lookup error in python autogenerated code

whaeck:
When trying to compile the python bindings, the "using njoy::GNDStk;" causes an ambiguous name lookup error for the Component aliasing. Fully specifying the namespace (and removing the using statements) resolves the error.

The generated python wrapper cpp files were updated as well.


staleyLANL (approved):
All looks fine. I may eventually make className() and related functions public, because it seems plausible that we may eventually have free functions - not member functions of Component - that do one thing or another with classes derived from Component. Such functions may want to access things like className() or keys(). If those change to being public at some point, I'd probably put the call to className() back into the autogenerated Python bindings, so that we have fewer literal strings floating around.

PR #41, opened 2021-06-08
Feature/v1.9 v4 - continued

whaeck:
Functional C++ unit tests for GNDS 1.9 nodes: axis, values

Functional python unit tests for GNDS 1.9 nodes: axis, values

Changes in json2class.exe for the python bindings:
  - the Defaulted class is not used for __init__ and replaced by the underlying type and a default argument (this avoids us having to wrap Defaulted)
  - setting default argument to std::nullopt for any metadata or child using std::optional
  - python names are now properly generated (e.g. valueType -> value_type)

In addition, I added the wrapper code for the Interpolation enum since I'll need this once I look at the Grid node.


autogen/json2class.cpp
```diff
@@ -1640,7 +1691,7 @@ void file_python_class(const NameDeps &obj, const std::string &filePythonCPP)
    cpp << "         python::init<\n";
    count = 0;
    for (auto &m : minfo) {
-      cpp << "            const " << m.fullVarType;
+      cpp << "            const " << (m.isDefaulted ? m.varType : m.fullVarType);
```
staleyLANL, line 1694: Is this because Python doesn't know about the C++ class Defaulted?

autogen/json2class.cpp
```diff
@@ -1650,34 +1701,59 @@ void file_python_class(const NameDeps &obj, const std::string &filePythonCPP)
    cpp << "         >(),\n";
 
    // python::arg...
-   for (auto &m : minfo)
-      cpp << "         python::arg(\"" << m.varName << "\"),\n";
-   for (auto &c : cinfo)
-      cpp << "         python::arg(\"" << c.varName << "\"),\n";
+   for (auto &m : minfo) {
```
staleyLANL, line 1704: I'd considered this autogeneration code to be sort of a hack, but I see it was clear enough (and/or you're smart enough) that you were able to figure out how to work with its data structures to make the changes you needed to make. Good work!

python/src/v1.9/transport/CrossSection.python.cpp
```diff
@@ -38,15 +39,15 @@ void wrapCrossSection(python::module &module)
             const std::optional<containers::Regions1d> &,
             const std::vector<VARIANT> &
          >(),
-         python::arg("XYs1d"),
-         python::arg("regions1d"),
+         python::arg("xys1d") = std::nullopt,
```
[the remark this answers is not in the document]
staleyLANL: Let me take a closer look at the issue you raised about the code seeming wrong. There's some hacky stuff in json2class, relating to places where you get a variant between two things. In the infoChildren struct, there's a flag called isChoice that sometimes needs to be checked, and what needs to be done in other places depends on the value. (Basically, is something just a regular child node, or is it combined with other choice child nodes into a single variant, or perhaps vector of variants.) The problem to which you refer may relate to this thing.

staleyLANL (approved):
Looks fine. I may make a few stylistic tweaks to what you wrote in json2class.cpp in order to keep the style consistent throughout with the original, but that's not really important. I'll mostly work on the other things we talked about. If I get a chance I'll look into that thing about XYs1d that I spoke about in a comment.

autogen/json2class.cpp
whaeck: Yes. It's one of the reasons I had you add those constructors not using Defaulted. The init here uses those.

autogen/json2class.cpp
whaeck: I'll take that as a complement ;-)

PR #42, opened 2021-06-09
In JSON, changed "link" "type" to "UTF8Text", not "bodyText".

staleyLANL:
That really seems better; what we call "body text" refers to something else.
New auto-generated Link.hpp and Link.python.cpp files reflect the above.
Also, no longer needed using bodyText = std::string in Support.hpp.
Which, again, could have created confusion with regards to the term.
Some cosmetic changes in json2class.cpp that I normally wouldn't have done...
...Except that I'm trying to keep some stylistic things consistent.
But I left in the new curly braces!
Importantly (not cosmetic), json2class deals w/body text for Python code.
Didn't quite get to the other things we talked about; those are next.

### DIFF
--- a/autogen/generalPurpose.json
+++ b/autogen/generalPurpose.json
@@ -11,7 +11,7 @@
                 "description": "The URL or xpath-like string pointing to the referred-to element",
                 "name": "href",
                 "required": false,
-                "type": "bodyText"
+                "type": "UTF8Text"
             }
         },
         "bodyText": null,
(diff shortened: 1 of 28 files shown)


staleyLANL:
The name feature/component-work isn't really quite right.
However, I'm also planning to put in changes to Component that we talked about.

whaeck (approved):
Looks good. I don't see why the name change for the help/documentation function is necessary, but it's not important anyway (an average user shouldn't even know about this function).

PR #43, opened 2021-06-10
Feature/getters and setters

staleyLANL:
Miscellaneous changes I've made mostly to the getter and setter code in (or for) the autogenerated classes.
There's more to do, but I'd like to get the present work out of the way.


autogen/json2class.cpp
```diff
@@ -1548,6 +1593,47 @@ void read(const std::string &file, nlohmann::json &jdoc)
       throw std::exception{};
    }
    ifs >> jdoc;
+
+   (void)firsttime; // so no "unused" warning
+#if 0
```
whaeck, line 1598: I'm not a fan of compiler directives for including pieces of code. You should consider putting this in an "information" function instead.

python/.gitignore
```diff
@@ -0,0 +1 @@
+GNDStk.so
```
whaeck, line 1: You should change this to *.so. GNDStk.so would be the name of the dynamic library if you compile with python 2.x, which is not what we want.

autogen/json2class.cpp
```diff
@@ -1483,7 +1528,7 @@ void make_class(
 
    // output: defaults (applicable only to metadata)
    oss << "\n   " << small;
-   oss << "\n   // Relevant defaults";
+   oss << "\n   // Relevant defaults;";
```
whaeck, line 1531: I'm nitpicking but you don't need a semi-colon here

whaeck (changes requested):
Nothing too important. We'll have to look at the setter functions using label and index in more detail. I'm not so sure about those yet but we can leave them in for now since I haven't tested those yet.

autogen/json2class.cpp
```diff
@@ -259,9 +259,8 @@ void check_class(
    assert(value.contains("attributes"));
    assert(value.contains("childNodes"));
    assert(value.contains("description"));
-   assert(value.contains("bodyText"));
```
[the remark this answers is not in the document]
staleyLANL: OK, I'll put it back in. I'd had the impression from our conversation the other day that perhaps we should allow it to be optional. No big deal.

autogen/json2class.cpp
staleyLANL: OK, will-do.

autogen/json2class.cpp
staleyLANL: I thought it "read" a bit better, but if you don't like it, I'll take it out. :-)

python/.gitignore
staleyLANL: OK, done.

whaeck:
@staleyLANL weren't you supposed to be off yesterday?

autogen/json2class.cpp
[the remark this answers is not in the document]
whaeck: Since these are asserts, it'll only be an issue when running in debug. In the future, we should refactor this as a separate tool to check the json file without generating code from it - we should do that once the prototype is ready.

whaeck (approved):

PR #44, opened 2021-06-13
Feature/v1.9 v4

whaeck:
There's enough in here already to merit a pull request, although it's a lot of small stuff.

First of all, a few minor changes in the autogeneration code:
- the custom.hpp files now have some pregenerated code as well (an empty help map used in python that we can then fill out)
- the grid element style attribute is an enumerator and is now treated as the other enumeratos like interpolation (there will be many like this)
- the variant in the case of choice is now printed properly
- when choice is used the appropriate argument in the python init function is printed

A few minor changes in the json files:
- the link element href attribute is now required (why have an element with only one attribute that is not even required?)
- changed the type of the grid style attribute to gridStyle to force the use of an enumerator

Unit test was added for the v1.9 Grid. The other v1.9 unit tests were slightly refactored to reflect the style of ENDFtk unit tests (focusing on creating components in various ways, testing their content and printing them out). The unit tests now use raw string literals to avoid having to use \" in XML snippets.

Changes on the python side:
- python bindings for the GridStyle enumerator since that is required for v1.9 Grid
- added a to_xml_string and to_xml_file function to print nodes

Outstanding issues discovered in the unit tests (see commented out code in Values.test.cpp, Grid.test.cpp and Test_GNDStk_v1_9_containers_Values.py):
- Debug mode compilation causes an assertion failure when executing the following code:
```
std::ostringstream out;
XML( Node( chunk ) ).write( out, false ); // chunk is a Values instance
```
- For Grid, the above code fails to compile (related to the toNode function)


formats/v1.9/readme
```diff
@@ -24,4 +24,23 @@ We're putting these into version control in order to track our changes, which
 we hope will be relatively small. (Except for some prettyprinting we'll do,
 through an editor, as the first change.) Substantive changes are geared towards
 the goal of helping our automatic code-generation tool to work more smoothly.
-Note that summary_gpdc.json is being renamed to summary_containers.json.
+
+The following changes have been made:
```
staleyLANL, line 28: Kudos for adding to my old readme. :-) That shows some dedication.

formats/v1.9/summary_containers.json
```diff
@@ -172,7 +172,7 @@
                 "default": null,
                 "description": "The URL or xpath-like string pointing to the referred-to element",
                 "name": "href",
-                "required": false,
+                "required": true,
```
staleyLANL, line 175: This file of course isn't used by the autogenerator yet, but I see you're putting things in order for when it is.

python/src/GNDStk.python.cpp
```diff
@@ -12,6 +12,7 @@ namespace core {
 
   void wrapNode( python::module& );
 
+  void wrapGridStyle( python::module& );
```
staleyLANL, line 15: And I know there are some other Standard Interface related enumerators floating around.

python/src/Interpolation.python.cpp
```diff
@@ -36,4 +36,4 @@ void wrapInterpolation( python::module& module ) {
   .value( "chargedparticle", Component::chargedparticle );
 }
 
-} // namespace rmatrix
+} // namespace core
```
staleyLANL, line 39: I wonder how this happened. :-D

python/src/v1.9/containers/Axes.python.cpp
```diff
@@ -36,12 +40,10 @@ void wrapAxes(python::module &module)
...
          python::init<
             const std::optional<UTF8Text> &,
-            const std::optional<std::vector<containers::Axis>> &,
-            const std::optional<std::vector<containers::Grid>> &
+            const std::vector<VARIANT> &
          >(),
          python::arg("href") = std::nullopt,
-         python::arg("axis") = std::nullopt,
-         python::arg("grid") = std::nullopt,
+         python::arg("choice"),
```
staleyLANL, line 46: By the way, I'm thinking of renaming those things "variant" rather than "choice"; the latter really isn't an ideal term, but was something I used just because it's used in the .json files. "variant" perhaps isn't ideal either, as it could be confused with std::variant. But it makes sense, on the other hand: variant is a std::variant. So, a user know what s/he's dealing with. Or, maybe there's a completely different name we should consider using.

python/test/Test_GNDStk_GridStyle.py
```diff
@@ -0,0 +1,86 @@
+# standard imports
```
staleyLANL, line 1: Some of this actually feels like highly redundant, over-the-top testing. :-) Like, it's true that value == none, and then it's false that value != none. But over-the-top testing is preferable to insufficient testing!

src/GNDStk/v1.9/containers/Axes/src/custom.hpp
```diff
@@ -0,0 +1,30 @@
...
+   *
+   +  Sort the elements by index so that the indices are in order
+   */
+  void construct() {
+
+    // useful lambdas
+    auto getIndex = [] ( const auto& variant )
+                       { return variant.index().value(); };
+
+    std::sort( this->content.choice.begin(), this->content.choice.end(),
```
staleyLANL, line 16: As I mentioned in our phone conversation, sorting by index is probably something I should do automatically. I think that the Component base class could be made to detect the situation when it exists.

src/GNDStk/v1.9/containers/Axes/src/custom.hpp
```diff
@@ -0,0 +1,30 @@
...
+                   { return std::visit( getIndex, left ) <
+                            std::visit( getIndex, right ); } );
+  }
+
+public:
+
+  /**
+   *  @brief Return the axis and grid elements defined in the axes element
+   */
+  decltype(auto) size() const {
```
staleyLANL, line 27: auto should be sufficient, as it's just going to be a std::size_t. By the way, it might make sense to have the autogenerator code make size() functions automatically, except it's presumably possible that a class it creates could have more than one vector in it, in which case we wouldn't know which one size() is for. So doing those in the customization seems like the way to go.

src/GNDStk/v1.9/containers/Axis/test/Axis.test.cpp
```diff
@@ -9,55 +9,66 @@
...
     WHEN( "the data is constructed from a node" ) {
 
-      Node core = chunk();
-      Axis chunk( core );
+      Node node;
+      node << chunk();
+      Axis chunk( node );
 
-      THEN( "an Axis can be constructed and members can be tested" ) {
+      THEN( "the component can be constructed and members can be tested" ) {
```
staleyLANL, line 50: I see how genericizing the terminology like this, will help in the long terms when we have lots of tests for lots of classes.

src/GNDStk/v1.9/containers/Axis/test/Axis.test.cpp
```diff
@@ -75,60 +86,46 @@ SCENARIO( "Axis" ) {
...
-}
-
-Node chunk() {
-
-  Node chunk;
-  chunk << "<axis index=\"1\" label=\"energy_in\" unit=\"eV\"/>";
-
-  return chunk;
+  return
+R"***(<axis index="1" label="energy_in" unit="eV" />
```
staleyLANL, line 127: I like this format for writing short, one-liner string literals. Easy to read.

staleyLANL (approved):
Looks fine, but perhaps see my comments before merging.

formats/v1.9/readme
whaeck: It'll be easier to remember what we changed :-)

formats/v1.9/summary_containers.json
whaeck: That's the general idea.

python/src/GNDStk.python.cpp
whaeck: I was thinking of adding all of them when I added this one but I'll do that once I'm done with all the other prototype testing. There was one in there (the encoding one) that I was wondering about what it was for.

python/src/Interpolation.python.cpp
whaeck: It's called a copy/paste and you know it ;-)

python/src/v1.9/containers/Axes.python.cpp
whaeck: I'm not a fan of choice either. We can leave it for now but we should change it before we generate a full standard compliant interface.

python/test/Test_GNDStk_GridStyle.py
whaeck: True, but the test is designed to verify the following:
- the enumerator class can be imported in python without error
- the == and != operators on the enumerator work as they should
- all possible values of the enumerator have been bound properly (if I had forgotten a specific value in the bindings, I would have gotten an import error or something similar when trying to use it)

It might indeed look silly, but you can't imagine how many errors I have found in my code by testing things that might seem obvious or over the top. I can generally ensure quality on code we write because of it.

src/GNDStk/v1.9/containers/Axes/src/custom.hpp
whaeck: Yes, I agree. This one can be generated automagically since we can deduce the situation.

src/GNDStk/v1.9/containers/Axes/src/custom.hpp
whaeck: In this case yes, auto would have been sufficient since it is an integral type. But I've been bitten before by inadvertently copying data by using auto as the return type instead of decltype(auto).

And again, this might indeed be a candidate for autogeneration.

src/GNDStk/v1.9/containers/Axis/test/Axis.test.cpp
whaeck: Indeed. I'm already thinking of at least autogenerating a skeleton unit test file once I have converged to a specific style.

src/GNDStk/v1.9/containers/Axis/test/Axis.test.cpp
whaeck: Even the larger ones are going to be clearer. And this way, we won't need escape characters all the time.

PR #45, opened 2021-06-22
Feature/values

staleyLANL:
All test codes appear to compile and run now, at least on my computer.

Take a look at the remarks I wrote into the Values custom.hpp file.

This PR involves substantial work on all of the code related to the BodyText class. At the moment, the code is a bit more wild and woolly, less refined, than I'd prefer; and, I have several "todo"s in there, for things that deserve a closer look. But I wanted to get this out there, because it appears to work. The next major step will be building a good set of tests for BodyText code.


src/GNDStk/BodyText/src/sync.hpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+
+// -----------------------------------------------------------------------------
+// pushToDerived(content)
+// Basically, the reverse of the above.
+// -----------------------------------------------------------------------------
+
+template<class CONTENT>
+void pushToDerived(CONTENT &content) const
```
whaeck, line 33: See above, I don't think the SFINAE works properly in these two functions due to the derived class being incomplete (same issue as the has_construct we have had to deal with).

whaeck (approved):
Made a few changes to get actions to work properly (you must have uncommented some of the wrapper function calls that had not been added to the CMakeLists.txt file while updating the auto-generated code).

The branch also showed conflicts because you copied the Values.test.cpp file instead of merging in the file from the current master which would have avoided the conflicts.

I think there's lingering SFINAE issues here similar to the has_construct issue we've encountered before. The ones I pointed out here are related to the BodyText (has_length, etc...) but the same thing will happen for the has_index and has_label that are being used in the getter functions.

I'm still approving it since I need these changes to continue working on the v1.9 prototype.

PR #46, opened 2021-06-23
Feature/v1.9 v4

whaeck:
Following the update: the latest state of the v1.9 prototype.

C++ unit test for XYs1d has been added, pieces of tests that used to fail are now uncommented since they work now.

On the python side: added Grid and Link, with unit tests.

Currently, only the get functions using index or label do not work as they should (see commented out test in the Axes class, an exception is thrown saying that the index or label is not present).


staleyLANL (approved):
Looks good. I'll look into the index/label issue.

PR #47, opened 2021-06-23
Feature/v1.9 v4

whaeck:
A few simple updates:
- restricted python properties based on choice and vector occurrence
- some changes in python bindings namespaces that interfered with proper namespace resolution for types (because the python wrapper functions were in namespaces similar to the component namespaces the compiler could not properly resolve component types)
- all python cpp files are now enabled (no more paying attention to commenting out containers and transport wrapper functions)

In short: the auto-generated python bindings now all compile!

Caveat: not all C++ and python unit tests are done but at least it compiles properly, and all C++ and python unit tests run properly

### DIFF
--- a/autogen/json2class.cpp
+++ b/autogen/json2class.cpp
@@ -1654,15 +1654,15 @@ void file_python_namespace(
 
    const std::string underver = replace(Version,'.','_');
    cpp << "// " << Version << " interface\n";
-   cpp << "namespace " << underver << " {\n\n";
+   cpp << "namespace python_" << underver << " {\n\n";
 
    cpp << "// " << nsname << " declarations\n";
-   cpp << "namespace " << nsname << " {\n";
+   cpp << "namespace python_" << nsname << " {\n";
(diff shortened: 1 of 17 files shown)


autogen/json2class.cpp
```diff
@@ -1671,18 +1671,18 @@ void file_python_namespace(
    cpp << "   // create the " << nsname << " submodule\n";
    cpp << "   python::module submodule = module.def_submodule(\n";
    cpp << "      \"" << nsname << "\",\n";
-   cpp << "      \"GNDS v1.9 " << nsname << "\"\n";
+   cpp << "      \"GNDS " << Version << " " << nsname << "\"\n";
```
staleyLANL, line 1674: I can't believe I missed that accidental hardcoding. :-/

whaeck: I just noticed that by chance.

staleyLANL (approved):
Look good. I'll trust all the python changes. Approving now...

PR #48, opened 2021-06-24
SFINAE shenanigans, traits trickery, and code consolidations.

staleyLANL:
The SFINAE issue with regards to querying a Component-derived class turned out to be unrelated to incomplete classes. There were, in fact, two issues. Consider, for example, length...

1. The derived class doesn't have data member length - it has data member content.length. It does have length *functions*, but it's unclear if the particular SFINAE construct would have known how to disambiguate overloads, and a failure to do so might have broken the SFINAE anyway.

2. Much more subtly, it turned out that CV (const/volatile) qualifiers, and a type actually being a *reference* (decltype(alternative) in a lambda), prevented the SFINAE from doing what it was intended to do. This is fixed with a remove_cvref.

The above issues have been addressed.

Commented-out code is Axes.test.cpp has been put back in, and works brilliantly.

The "todo" about pointers is something we should talk about. There's actually a reason I chose to return pointers from certain very specific autogenerated getters. Let's have a conversation about this. In any event, for now, I commented-in that code as well, and dereferenced things when I needed to. Works brilliantly.

Throughout the code base, I reduced code bulk by making certain changes in the use of traits classes. (This is something I've been wanting to do for a while.) For example, enable_if can be replaced with enable_if_t; this allows us to remove "typename" at the beginning and "::type" at the end. Likewise, is_same can be replaced with is_same_v, allowing "::value" at the end to be removed. The reduced clutter makes the code more readable, without, in my opinion, making the code any less understandable. If anything it's more understandable, because of less clutter. These changes might be considered for other NJOY codes.

I did the above with several Standard Library classes, and in a few cases I changed GNDStk classes to allow for the same pattern. (Like, with those traits classes that previously had SFINAE issues.) I could do a bit more in this respect, but want to get to other things now.

Before, I'd given a certain GNDStk traits class (is_oneof) a double meaning: Is the type one of the variant alternatives, or the variant itself. For what it was originally designed for, the double meaning gave me what I needed. As I began using this class in more places, however, I sometimes had to add additional code to cancel-out the second meaning. I've now redesigned all that stuff so that meanings are more clear, and not mixed together.

Todid some todos.

I changed is_arithmetic to isAlternative (e.g., one of the vector alternatives in the variant) in the Values customization. That way, any type that should work, will work. In particular, I don't think std::string - one of the types I support - would have worked before.

Some near-identical pairs of functions regarding (1) index getters, and (2) label getters, have been consolidated. This saved a bit of code bulk. The replacement code is very slightly longer, but it's just one function where two existed before.


whaeck:
The CI seems to fail on MacOS 10.15 using AppleClang 12, which is probably the same configuration as my current work laptop (MacOS 10.15.7 with AppleClang 12). I've been able to reproduce the error with this branch on my laptop (it seems to happen only for -DCMAKE_BUILD_TYPE=Debug).

There appears to be an issue in the Values unit test, in particular when transforming the Values instance to a Node to write it out as an XML string.
```
-------------------------------------------------------------------------------
Scenario: Values
     Given: valid data for the component
      When: the data is constructed from a node
      Then: it can be written in XML
-------------------------------------------------------------------------------
/Users/wim/Projects/njoy21/GNDStk/src/GNDStk/v1.9/containers/Values/test/Values.test.cpp:76
...............................................................................

/Users/wim/Projects/njoy21/GNDStk/src/GNDStk/v1.9/containers/Values/test/Values.test.cpp:81: FAILED:
  CHECK( out.str() == string )
with expansion:
  "<values length="4" start="0" valueType="Float64">2500 8.9172 2550</values>
  "
  ==
  "<values length="4" start="0" valueType="Float64">2500 8.9172 2550 8.9155</
  values>
  "
```
For some reason, the last value (8.9155) does not get printed out.

The weirdest thing about it, is that it does not fail every single time I execute the test. It looks like a 50 % failure rate.

whaeck:
In addition to the above, the changes also lead to the following warning messages being output over and over again when running tests that have a Values node in them:
```
[info]    Re-forming vector of one type into vector of another type;
          was this intentional?
```
Prior to the change, these information messages were not present. It appears to happen when the Values node is constructed from a Node. Contrary to the previous comment, this seems to happen for every OS, compiler and compilation mode.

whaeck:
I managed to reproduce both problems (Debug Values test failing 50% of the time and the information messages appearing) with minimal changes starting from the feature/v1.9-v4 branch. I pushed these changes on the test/cleanup branch.

Essentially it seems to be introduced when the following changes are made:
- changing sfinae in details.hpp in BodyText and Component (changing index to content.index, etc)
- decaying/removing the cv qualifiers in the BodyText sync.hpp file and the Component detail.hpp (I used std::decay_t<> but the remove_cv_t<> does the same thing).
See changes here: https://github.com/njoy/GNDStk/commit/0f9298878f955152c76b7dd20c02d00cc19f42fa

staleyLANL:
I put in some print statements (just locally, not in the uploaded code) that verify that the SFINAE, both for pull and push, is working now.

staleyLANL:
Remark about the classes for SFINAE in the two `detail.hpp` files we spoke of:

Component `detail.hpp` defines `hasIndex` and `hasLabel`; these are helpers for autogenerated classes with index- and label-based getters.

BodyText `detail.hpp` defines `hasLength`, `hasStart`, and `hasValueType`. These are designed similarly to the other two, but deal with different specific fields.

I should probably eventually put all five of these in one place, and write them in a slightly more consistent manner, but they do all have different purposes.

whaeck:
Seems like the changes solved the 50% failure rate too. I'm not seeing it happen anymore on my laptop.

staleyLANL:
> Seems like the changes solved the 50% failure rate too. I'm not seeing it happen anymore on my laptop.

Yeah, that was because of the off-by-one. The missing -1 in getBounds. I seldom make those nowadays, but I guess something gets through every once in a while.

staleyLANL:
> The remaining issues seem to have gone away. With the exception of the pointer for the index and label based lookup functions, it looks fine.
> 
> One remark though: your naming scheme for type names is getting a bit erratic. For the recent SFINAE naming/renaming, I've seen various ways: like_this or likeThis and even a few LikeThis for ome of the older SFINAE.

One reason for things like `has_length` and `hasLength` is that the first is a helper for the second; so, they need different names. So, that choice was intentional. I suppose I could follow the system they use in `std::`, and write `has_length` and `has_length_v`. Then again, NJOY's system is already different from that in `std::`. Again, I'm not worrying too much about it in `detail::` stuff. It can always be revisited later if there's time.

PR #49, opened 2021-07-01
Completion of test suite for class BodyText

staleyLANL:


src/GNDStk/BodyText.hpp
```diff
@@ -57,56 +45,74 @@ class BodyText {
...
 
-   // Get length, start, and valueType from derived-class content,
-   // or possibly put them there (if we determined them from a vector)
+   // Pull/push length/start/valueType from/to derived-class struct content
    #include "GNDStk/BodyText/src/sync.hpp"
 
    // Assignment
-   // From vector or string; the latter == calling our raw string setter
+   // From string or vector; the former == calling our raw string setter
    #include "GNDStk/BodyText/src/assign.hpp"
```
whaeck, line 116: Having a BodyText constructor for either vector<T> and string similar to the assignment operator would be appreciated. It'd simplify some of the custom constructors.

src/GNDStk/BodyText/src/types.hpp
```diff
@@ -0,0 +1,35 @@
...
+   std::vector<unsigned char>,
+   std::vector<unsigned short>,
+   std::vector<unsigned int>,
+   std::vector<unsigned long>,
+   std::vector<unsigned long long>,
+
+   // floating-points
+   std::vector<float>,
+   std::vector<double>,
+   std::vector<long double>
```
whaeck, line 31: I don't think there's a need to support so many different types. string, int and double will be more than sufficient.

src/GNDStk/BodyText/test/types.test.cpp
```diff
@@ -0,0 +1,88 @@
...
+      CHECK(std::variant_size_v<vv> == 15);
+      vs scalars;
+
+      {
+         vectors = std::vector<std::string>(10,"a");
+         scalars = std::get<std::vector<std::string>>(vectors)[0];
+         CHECK(std::get<std::string>(scalars) == "a");
+      } {
+         vectors = std::vector<char>(10,'b');
+         scalars = std::get<std::vector<char>>(vectors)[0];
```
whaeck, line 31: same as above

whaeck (changes requested):

src/GNDStk/BodyText.hpp
```diff
@@ -2,10 +2,10 @@
...
 // When writing a Component with its generic write() function (or its stream
 // output, which uses write()), and the Component is based on a BodyText<true>,
-// values will be printed in GNDStk::columns columns. "columns" is aliased to
+// values will be printed with GNDStk::columns across. "columns" is aliased to
 // "across" for convenience, because, at the time of this writing, GNDStk has
 // a Meta<> object, named "columns", which would also be in scope if the core
-// namespace is used. So, a user might prefer to use "across".
+// namespace is used. So, a user might prefer to use the name "across".
 inline std::size_t columns = 4;
 inline std::size_t &across = columns;
```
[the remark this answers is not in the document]
staleyLANL: My I/O system is a good one, and very carefully thought out. What I perhaps should do, at some point, is to consolidate all printing-related parameters into their own sub-namespace within `GNDStk::`, and put in one place. At the moment, some of the printing-related parameters were added recently, because they related to new features, and were placed where it was convenient to do so. But I can look at this at some point, locate anything in this regard, and consolidate it all in one place. They, I think it'll be clearer why I'm doing things this way.

GNDStk's overall I/O system, and in particular its printing capabilities, applies to at least the following: print XML, print JSON, print a Node in an internal format that's useful for debugging, and print objects that are derived from Component. Things like `GNDStk::indent` are sort of like environment variables in linux: set it to what you like, and then it applies to whatever uses it, so you no longer need to clutter individual print commands with stuff like that. Set it once, instead of many times.

C++ does not, at present, allow additional arguments to things like `operator<<`. Even if it did, using it in such a way would probably require that someone uses the awkward functional form of the call; something like `std::cout.operator<<(value,then,some,additional,arguments)`. Instead of something nice and short like `std::cout << value`, with "environment variables" set up beforehand, the way you like them, and automatically used in any `<<` invocation that happens to respect the environment variable. Coloring (for errors, warnings, and the like; and also for nice color-coded output of objects derived from Component), and even whether or not coloring is switched on, is also something for which these convenient global settings can be made.

We all learn early in our computer science classes that "global variables are bad." :-) For many things, they are. However, also like many things, when used judiciously they can be a good thing. Most things are - appropriately - function parameters. However, over the years I've developed a nice I/O system that allow certain niceties (basically, pretty-printing issues like indentation and colorization) to be set up with namespace-scope variables (namespaces are made for polluting!!), to set user preferences that they probably want *globally*. Then, individual prints and short and sweet. With several parameters available, default argument lists (if they were even possible with things like `operator<<`) might require that several "defaults" be given explicitly in order to get to one default, later in the argument list, that someone wants to change. (Because as you know, C++ isn't flexible with argument ordering, as Python is.)

Things that should still be parameters, are parameters. An example is indentation level. (Not to be confused with GNDStk::indent, which is the number of spaces you like, PER level. I like 3, you prefer 2.) For example, `Component::write()` (which automatically imbues its derived class with a sophisticated, customizable, and colorizable print capability) has a level that defaults to 0. It also provided an `operator<<`, which doesn't (and can't, per the C++ standard) allow level to be set. So, `operator<<` calls `write()` with `level=0`. Once inside the `write()` ecosystem, the code calls other `write()` functions, with each dealing with `level` in a nice way. The result will be, for example, that users will have a SUPER nice and flexible printing capability for absolutely *every* GNDS Standard Interface class - all those things that derive from Component. That is, Component provides very generic pretty-printing capabilities, and the "environment variables" that control stylistic aspects of printing are available, separately, for anyone who wishes to use them. Users get a whole lot, for relatively little effort on our part.

This general system is something I've developed over the years, and it's very easy-to-use, effective, and substantially better than other alternatives I've seen.

src/GNDStk/BodyText/test/BodyText.test.cpp
```diff
@@ -0,0 +1,93 @@
...
+   } {
+      BodyText<true> nbt;
+      CHECK(nbt.length   () == 0);
+      CHECK(nbt.size     () == 0);
+      CHECK(nbt.valueType() == "");
+      CHECK(nbt.string   () == "");
+   } {
+      BodyText<false> nbf;
+      // no data for <false>
+   }
```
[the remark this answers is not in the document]
staleyLANL: Weird is in the eye of the beholder. :-) But, yes, I can do this. Would you be fine if I just do these things for the next PR, which will include Component tests as well? Given that I've already tweaked things here and there in BodyText's code base, which of course is related to Component, this would be less messy than dealing wtih a merge.

src/GNDStk/BodyText/test/detail.test.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+
+   // ------------------------
+   // getBounds
+   // ------------------------
+
+   GIVEN("Testing detail::getBounds") {
+
+      // Various vector<int>s
+      {
```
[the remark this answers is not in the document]
staleyLANL: OK.

src/GNDStk/BodyText/test/BodyText.test.cpp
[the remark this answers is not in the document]
staleyLANL: I've done like this in a handful of other test files from before, and can change it in those places as well, to be consistent.

src/GNDStk/BodyText.hpp
staleyLANL: BodyText actually has no explicitly defined constructors. And, Component (which of course derives from BodyText, and *from* which our Standard Interface classes are derived) has only one constructor - one that must receive the derived-class fields that it links with the multi-query. The way things are set up with these classes, I think we're talking here about `finish()` functions, not so much constructors proper. (Issues involving order of initialization during construction have to be considered here, and are the root cause behind the need for the `finish()` functions in the first place.) I do have a `finish(vector)` function available now. Maybe there's a need for a `finish(string)`. Let me think about this.

src/GNDStk/BodyText/src/types.hpp
staleyLANL: Probably not, although it's easy to support them, and I'm inclined to think that an occasional user may actually prefer certain specific types other than string, int, and double. I'd want to support at least float and long double, in additional to double, as I can see scenarios where either of those might be warranted.

src/GNDStk/BodyText.hpp
staleyLANL: One thing I think we agreed on, in a discussion a little while back, is that we'll work for now on the assumption that length and start are parameters for saving space (by allowing for zeros to be removed) in GNDS files. But, when we're storing and working with a vector, we'll say that all values (including leading and trailing 0s) are actually in the vector, in memory. This saves us from dealing with extra complexities in the BodyText class, and from ambiguities such as what, say, vector element `n` would mean if `start` is nonzero. Would a user want the actual thing in element `[n]` of the vector, if they asked for it, or would they really want `[start+n]`? Anyway, the way the class works, the vector has any necessary zeros, and start should be considered to be 0. The code will automatically deal with leading and trailing zeros when dealing with the string form.

This probably has ramifications for what's in the Values `custom.hpp` at the moment. For the construction from vector, we should say it's a vector with all elements, and dispense with `length`, `start`, and `valueType`.

src/GNDStk/BodyText/test/types.test.cpp
staleyLANL: I've made these changes, in the files you mentioned here, and a handful of others in the code base. Changes are in my newest branch, which I'll upload shortly.

staleyLANL:
I've made changes regarding things like GIVEN/WHEN/THEN in the test suite. These are actually in my newer branch, which also has tests for nearly everything in Component. (I'm not testing its sort() directly yet, but it's indirectly tested in various other places. A direct test requires lots of permutations, given how flexible the sort() is, and will take some more time.) Perhaps this current PR should just be withdrawn, and then I'll put in another one - one that includes the new work as well.

whaeck (approved):
Some of the requested changes will be done later.

PR #51, opened 2021-08-03
Feature/misc

staleyLANL:
This PR contains lots of miscellaneous minor changes that I've buffered up over the past few weeks. Sorry for the apparent lack of specificity here. :-) Basically, I just want to get these out of the way, as I'll be uploading work regarding floating-point precision soon, and would like to do that by itself, not cluttered by this material.


src/GNDStk/v1.9/containers/Values.hpp
```diff
@@ -81,21 +81,27 @@ class Values : public Component<Values,true> {
...
+   start() const
     { return content.start.value(); }
-   auto &start()
+   Integer32
+   start()
     { return content.start.value(); }
 
    // valueType
-   const auto &valueType() const
+   const UTF8Text
```
whaeck, line 100: I assume that UTF8Text is a std::string?

I think you're missing a reference '&' in the return type of this function and the next one. The other functions of this type seem to include the proper character, these seem to be the only exception so far.

src/GNDStk/v1.9/containers/XYs1d.hpp
```diff
@@ -96,39 +96,51 @@ class XYs1d : public Component<XYs1d> {
...
 
    // interpolation
-   const auto &interpolation() const
+   const enums::Interpolation
+   interpolation() const
     { return content.interpolation.value(); }
-   auto &interpolation()
+   enums::Interpolation
+   interpolation()
     { return content.interpolation.value(); }
```
whaeck, line 112: I was wrong, here's another one where you're missing the reference.

src/GNDStk/v1.9/key.hpp
```diff
@@ -0,0 +1,119 @@
+
+// THIS FILE WAS AUTO-GENERATED!
+// DO NOT MODIFY!
+
+// This file contains Meta and Child objects for the current GNDS version.
+// If someone wishes to use the GNDS Core Interface with the classes that
+// we've autogenerated in this context, these should prove to be useful.
+
```
whaeck, line 8: Does this mean we can remove the definition of the child and meta key names in the core interface and add them to the autogenerated code instead? It makes sense to have these only provided by GNDS standard version.

src/GNDStk/v1.9/transport/Reaction.hpp
```diff
@@ -88,27 +88,35 @@ class Reaction : public Component<Reaction> {
    // ------------------------
 
    // ENDF_MT
-   const auto &ENDF_MT() const
+   const Integer32 &
+   ENDF_MT() const
     { return content.ENDF_MT; }
-   auto &ENDF_MT()
+   Integer32 &
+   ENDF_MT()
```
whaeck, line 95: I'm confused here. When returning the Values node start value, the return types there were const Integer32 and Integer32 (without the reference). Is that because start needs to be treated differently from these (because they're BodyText related)?

src/GNDStk/Child/src/operator.hpp
```diff
@@ -284,3 +284,53 @@ inline auto operator+(
...
+
+// -----------------------------------------------------------------------------
+// Child || Child
+// Create a Child that represents one or the other of two children
+// -----------------------------------------------------------------------------
+
+// <void> || <void>
+// Return type is Child<void,...>
+template<
+   Allow ALLOW, class CONVERTER, class FILTER
```
whaeck, line 298: I really like how this now allows for a better syntax when specifying a variant child.

I've never liked the following syntax:
```
++Child<>("axis grid")
```
since it does not really convey the meaning of "look for many 'axis' or 'grid' nodes". Using a single string to store all variant type names is also not the best idea (what if I mistype and forget the space).

The new syntax like this:
```
++(Child<>("axis") || Child<>("grid"))
```
does a much better job at conveying this meaning properly. I like it.

whaeck (approved):
Changes to be made later:
- allow "__class__" and "class" and "__namespace__" and "namespace" for backwards compatibility reasons
- add the "variant" key to specify the variant name for a "choice+" or "choice" instead of "choice+ foo"
- use the "variant" value to name the variant type instead of for instance LINK_VALUES
- maybe change return type for getter/setter? To see what "looks" better.

src/GNDStk/v1.9/containers/Values.hpp
staleyLANL: `UTF8Text` is aliased to `std::string`. I use the former only because it's what they say in the JSON files. I do the same with regards to `Integer32` and `Float64`. The alternative would be to do more by-hand modifications, to the JSON files, than what we already do.

The inconsistencies in terms of constness and reference-ness, in places that use `Defaulted`, is due to the issue we spoke about in person. Everything can (and in my opinion should) be made consistent in the autogenerated files, if we give Python a correct understanding of class `Defaulted`, in the same fashion that Python understands `std::optional`. Without that fundamental fix, we'll always need to use `value()` in `Defaulted` objects, and getting correct semantics everywhere will be a game of whack-a-mole. So let's think in terms of doing something with Python.

src/GNDStk/v1.9/containers/XYs1d.hpp
staleyLANL: Again, this is because `interpolation` is a `Defaulted`. We need to give Python the correct understanding of `Defaulted`, and then can get rid of the `.value()` and make everything consistent.

src/GNDStk/v1.9/key.hpp
staleyLANL: Yes, it means that, and is a good thing. The `Child` and `Meta` objects in the present Core Interface were something I put in, early on, based on what I saw in the existing GNDS files. In other words, they were sort of ad-hoc and probably incomplete, but necessary in order to write functioning codes. Now, with the autogenerator, we can (and should) make them automatically, with one complete set of such objects for each GNDS version.

Right now, the difficulty of removing the original ones is that many (probably most!) of the codes in the test suite depend on these old `Child` and `Meta` objects. So, phasing out the old ones requires planning, to avoid a spectacular breaking of the test system. A good solution might be to put them into, say, a `test` or `testing` namespace that would be excluded from the Core Interface, but kept around indefinitely (probably as an undocumented feature, so as not to confuse users by its existence in the code), for use with the tests.

src/GNDStk/v1.9/containers/Values.hpp
staleyLANL: Another option would be to have the autogenerator automatically recognize `Integer32` as `int`, and so forth, so that the type aliases aren't even needed in GNDStk. A question would be whether the GNDS standard will ever *really* do something that makes the "32" and "64" and "UTF8" important in some way. Languages such as XML are very careful to define how they encode things in a very precise way. For GNDS, I'm guessing that it's not really important that an `Integer32` go into a 32-bit integer; if integers are 64 bits instead, that should obviously be OK. Doubles are typically 64 bits on modern machines, but don't need to be. "UTF8" is a specific thing, but I'm guessing that a C++ `std::string` will suffice for whatever someone puts into a GNDS "UTF8" string. We should probably have a discussion sometime about whether the GNDS standard's stated specificity, with these things, is actually very important or not.

src/GNDStk/Child/src/operator.hpp
staleyLANL: It's worth noting that you can still use the "name1 name2 ..." construct, in a single child, to get this effect, if you really want to do that. The `||` operator just does that for you. I *could* make `||` create some sort of special object, like perhaps a `tuple` that contains two or more `Child` objects. When I originally implemented all this (it was for handling `variant`), I chose to use `Child` as-is, with the spaces between names, because the alternative of using a special object, like a `tuple`, would require going through the entire code base and putting in additional overloads of all sort of functions that current accept only a `Child`. So, using `Child` under-to-hood serves as a minimally-disruptive implementation, and then the `||` operator is a nicety for users.

staleyLANL:
I've made a note of the "Changes to be made later," and will make them later. :-)

PR #53, opened 2021-08-10
Feature/autogen namespace etc

staleyLANL:
I decided to do a small, easy PR for this one.

This is one of the (several) changes we spoke about to make the code autogenerator more flexible. Here, we'll just allow the underscores around namespace and class, or allow them to not be there.

I left the prototype JSON specs the same, so they can serve as examples for other things, but changed the formats/v1.9 JSON files back to having the underscores.


whaeck (approved):
I fixed the failure with the python interface in the CI, using a lambda to return the underlying value in the Defaulted class. This is now the way it is coded in json2class.

In addition, the change in a component constructor with a Defaulted as metadata was not ported to the python interface. This has been corrected in json2class as well.

PR #56, opened 2021-10-12
Feature/new autogenerator

staleyLANL:
It's probably time to put this in, because I'm already working on some new things.

The new code autogenerator completely replaces the old one now.

The new one isn't perfect yet - there are still some things I want to do with it - but it's much better than the old one. The handling of regular child nodes versus that of "choice" child nodes is no longer tangled up, for instance. Data structures and functions are laid out in a much clearer and more logical fashion. A small but carefully designed "writer" class helps make code output much more clear. Basically, this autogenerator code is a major advance over the original.


whaeck (approved):
The new autogenerator looks a lot better and is better structured. The layout of the data is more clear and easier to use. I like it.

I merged the v1.9-v5 in this branch as well and made sure nothing changed in the generated files, except for the changes on the python side that I wanted. No files (except the python cpp files) changed with respect to what was already there.

PR #57, opened 2021-10-19
Feature/type cleanup

staleyLANL:


whaeck (approved):
Mostly cosmetic changes to the files (replacing GNDS speak for types to C++ fundamental type names). Autogenerator still produces the same files so nothing to remark here.

PR #58, opened 2021-10-22
Feature/v1.9 v5

whaeck:
A few additions from the v1.9 work to be added before we add the new auto-generator:
- Previously, the python auto-generation code was written with a custom constructor in mind. This is removed in favour of the auto-generated ones. Tests are cleaned up as a result.
- Adding UTF8Text in body text code
- Adding testing for Regions1D

I'll update the new auto-generation code to take the change on the python side into account as well.


whaeck:
I'm closing this one since these changes are now integrated in the feature/new-autogenerator branch

PR #59, opened 2021-12-10
Feature/hdf5

staleyLANL:
Simplistic (type-wise) but functional HDF5 capability.
Giving this PR a shot right now. Might as well get it out of the way.
Well, this has already been more than a 40 hour week. :-/


autogen/.gitignore
```diff
@@ -1 +1,2 @@
 json2class.exe
+test
```
whaeck, line 2: Why are you adding the test directory to be ignored by git?

autogen/v1.9.json
```diff
@@ -1,4 +1,7 @@
 {
+    "GNDSDir": "test",
```
whaeck, line 2: Shouldn't this be "Path" now instead of "GNDSDir"?

python/src/v1.9/transport/ReactionSuite.python.cpp
```diff
@@ -88,23 +88,6 @@ void wrapReactionSuite(python::module &module)
...
-
-          return Component( tree( child::reactionSuite ) );
-        },
-        python::arg( "filename" ),
-        "Read a reaction suite from an XML or json file\n\n"
-        "An exception is raised if something goes wrong while reading the\n"
-        "component\n\n"
-        "Arguments:\n"
-        "    filename    the name of the file"
-      )
```
whaeck, line 107: You should put this back.

src/GNDStk/JSON/src/write.hpp
```diff
@@ -9,7 +9,7 @@
 
 std::ostream &write(std::ostream &os, const bool decl = true) const
 {
-   (void)decl; // ignored
+   (void)decl; // unused, at least for now
```
whaeck, line 12: If decl is not used but you want the signature to remain the same, just remove the name. In that case, you don't have to add dummy code to get the argument to be "used":
```
std::ostream &write(std::ostream &os, const bool = true) const
```

I looked it up, it's legal.

src/GNDStk/Node/src/write.hpp
```diff
@@ -66,10 +69,19 @@ std::ostream &write(std::ostream &os, const int level) const
...
 // For FileType::text
 // -----------------------------------------------------------------------------
 
+private:
+
+// fixme I noticed that this write() variant isn't used (other write() functions
+// with filename go through ostream first, and call the earlier write() helper
+// with const int level). Decide if there's any reason to keep it. If we do keep
+// it, then it needs to be exercised in the test suite, which it isn't now.
+#if 0
```
whaeck, line 83: I'm not a fan of preprocessor directives to comment out code.

src/GNDStk/Node/src/write.hpp
```diff
@@ -12,17 +12,20 @@
...
 
 
 
 // -----------------------------------------------------------------------------
+// Helper
 // write(ostream, int level)
 // For FileType::text
 // -----------------------------------------------------------------------------
 
+private:
```
whaeck, line 27: This is not very important since this is only used for pretty-printing nodes, but why did you declare some functions private and other public?

src/GNDStk/convert/src/HDF5.hpp
```diff
@@ -0,0 +1,188 @@
...
+   static const std::string context = "convert(Node,HDF5)";
+   try {
+
+      // Open temporary HDF5 file. This *should* work, but we check anyway.
+      std::ofstream ofs(h.filename, std::ios::binary);
+      if (!ofs) {
+         log::error("Could not open temporary HDF5 file \"{}\"", h.filename);
+         throw std::exception{};
+      }
+      ofs.close();
```
whaeck, line 28: I would assume the HDF5 library to throw an exception if this happens?

src/GNDStk/v1.9/transport/CrossSection.hpp
```diff
@@ -149,6 +149,10 @@ class CrossSection : public Component<CrossSection> {
       XYs1d_regions1d(label) = obj; return *this;
    }
 
+   // XYs1d_regions1d(value) for vector push_back
+   CrossSection &XYs1d_regions1d(const XYs1d_regions1d_t &obj)
+      { setter(XYs1d_regions1d(), obj); return *this; }
```
whaeck, line 154: I have my reservations on adding this type of setter function. Some of the requirements on GNDS are that specific attributes like label and index are unique. This might potentially break that.

autogen/json2class.cpp
```diff
@@ -1255,48 +1266,23 @@ void writeClassCtorBody(writer &out, const std::string &argName)
...
-   out(1,"// from node");
-   out(1,"@(const Node &node) :", per.clname);
-   writeClassCtorComponent(out, per, false);
-   out();
-   writeClassCtorBody(out, "node");
-
    // ------------------------
-   // ctor: fields
+   // ctor: default,
+   // and from fields
```
whaeck, line 1271: Merging the default and from fields constructor, I like it ;-)

src/GNDStk/HDF5/test/HDF5.test.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+// -----------------------------------------------------------------------------
+// SCENARIO
+// -----------------------------------------------------------------------------
+
+SCENARIO("Testing GNDStk HDF5") {
+
+   /// fixme adapt code from the corresponding JSON test
+
+}
```
whaeck, line 17: I assume you are planning on adding tests?

whaeck:
I have changed the pull request to merge into the develop branch instead of the master branch so that we can keep the master branch clean until we are ready to do a release. That way, the master branch will always point to an official release. This is in line with what is done on libraries like ENDFtk and ACEtk.

whaeck (changes requested):
There are no tests for the new HDF5 capability - I assume these were added in the other pull requests?

Some code on the python side was inadvertently removed (the from_file function, see the remark on that). It should be added back. While I do have some other remarks, I think this is practically good to go.

staleyLANL:
> I have changed the pull request to merge into the develop branch instead of the master branch so that we can keep the master branch clean until we are ready to do a release. That way, the master branch will always point to an official release. This is in line with what is done on libraries like ENDFtk and ACEtk.

OK.

staleyLANL:
This was an initial HDF5 capability, but one I wanted to get out of the way; hence the PR. Lots more has been done with HDF5 since I put this in, and in fact, there's still a bit more to do with it. This original HDF5 capability just wrote everything as strings, similarly to how it's stored internally; obviously, one improvement is to actually write things like ints and doubles to HDF5 as...well, as ints and doubles. Some other refinement was needed with HDF5; for example, writing things like "CDATA" and "PCDATA" (as they're called in XML) in a good way.

Because I knew I'd be doing lots more work with HDF5, I held off on writing tests, which I knew would need to be changed anyway when I improved the HDF5 handling. In fact, it's just recently (this week) that I started putting things into the main HDF5 test code. So, no worries, tests are on the way; just not in this old PR.

src/GNDStk/HDF5/test/HDF5.test.cpp
staleyLANL: Yep; see other comment.

autogen/json2class.cpp
staleyLANL: Yeah; I'd initially had the code generator write a default constructor, and then various other kinds of constructors. After I settled on what constructors it should write, I realized that those two could be merged. And the code is shorter and probably more clear.

src/GNDStk/convert/src/HDF5.hpp
staleyLANL: I didn't actually check to see precisely what the HDF5 library does. Either way, I'd have my own "try to open it," so that the very first error message that's printed, if the open fails, is GNDStk's. This way, the look and feel of the HDF5 interface is the same as the look and feel of our XML and JSON interfaces. Generally speaking, GNDStk's XML, JSON, and HDF5 classes are designed to serve as lightweight wrappers around the classes in pugixml, nlohmann json, and highfive, and we do this precisely so that all the behavior can be given the same look and feel. The way I deal with diagnostics is just one aspect of this goal.

src/GNDStk/v1.9/transport/CrossSection.hpp
staleyLANL: Since you and I last spoke in detail about the code that the code generator builds, I've made various tweaks to the getters and setters, based on my experience in using it for Tom's work.

Something that came up early was that doing things with generated classes proved to be more syntactically messy than I'd anticipated. One would need to write, for example, `()` to get an underlying vector, before doing a `push_back()`. With optionals, a `*` or `->` was needed, and those would fail if there was nothing in the optional.

I made various improvements so that the use of the generated classes would be much smoother and more natural. For example, imagine wanting to push a new value into an optional vector that currently had no value. (Meaning the optional had nothing yet, not that it had a vector with no elements.) Doing so would have been very confusing to users, so I wrote functions -- like the one you're looking at -- precisely to give users a simple way of doing things.

In terms of your concern about the requirements of GNDS, two remarks...

First, I'd note that our primary goal is of course to support GNDS. However, an important secondary goal is to allow other non-GNDS data structures to be created. With that secondary goal in mind, some of what I've done in the various open pull requests is geared towards saying, essentially, "let's make the code generator more general, so that it doesn't impose GNDS-isms on anyone." It's possible, for example, that someone *might* want to design classes that have non-unique label or index. Jane User's `label` or `index` might have a completely different meaning that their meaning in GNDS.

Second, the concern about unique GNDS `index` and `label` is certainly valid, but I think that the issue comes up anyway, no matter what we do. Whatever mechanism we provide for adding vector elements, somebody might try to add two elements that both have the same `index` or `label`, if those fields exist in the element type. What you're seeing here is a convenient shorthand, avoiding cluttered syntax, for adding an element. Guaranteeing the uniqueness you speak of is really a different issue.

src/GNDStk/Node/src/write.hpp
staleyLANL: Because I wanted some to be private, and others public?! :-D

Seriously.... The ones that I declared public are the ones I want users to see, whereas the private ones are really just intended to be helpers for the public ones.

I suppose I could put the helpers into `detail::` functions instead, if anyone feels really strongly about this. Here, it was convenient to put them into actual member functions, so that I get the implicit `*this`, and also they're close to the `write()` functions that use them.

autogen/.gitignore
staleyLANL: Ah, this points out something that I hadn't even thought about. At some early point in the development of the code generator, I wanted a "sandbox" area to try out various schemes for autogenerating directories and files. I called my sandbox area `test/`, not really thinking about the fact that we have `test/` directories, with a specific purpose, throughout GNDStk.

`git status` would always mention my `test/` sandbox, so I put `test` in `.gitignore` in order to quiet those messages.

Also, at some point, I placed a symlink for `json2class.exe` in my own `autogen/` directory, pointing to the `json2class.exe` in the `build/` directory, for convenience.

I no longer have that symlink, and don't use `test/` any longer. So, for the repo, I should just get rid of that `.gitignore` file entirely. It serves no purpose in terms of what's in the repo itself. I'll make this change in my next commit.

autogen/v1.9.json
staleyLANL: Yes; I'll change it.

Note that we're actually using prototype.json, not v1.9.json, to generate the small "GNDS v1.9" prototype codes that are currently in the repo. Therefore, we should currently only be running `json2class.exe prototype.json`. The `v1.9.json` file is only going to be relevant if and when we fix up the weirdnesses in all of those `summary_*.json` files, and then also write tests and such for all of those. I anticipate that we may eventually want to refactor the `autogen/` contents, so that our prototype stuff appears, in some form, in a proper `autogen` `test/` directory. Then, we may remove the "v1.9" business entirely and go directly to GNDS 2.0. This is all another discussion for another time.

python/src/v1.9/transport/ReactionSuite.python.cpp
staleyLANL: I did all this work a while back, and don't actually recall *why* this stuff disappeared. I don't remember taking it out, but if I did, I no doubt had what I thought was a good reason for doing so. Let me look into this.

src/GNDStk/Node/src/write.hpp
staleyLANL: Sometimes I'll use an `#if 0`, as opposed to a block comment (or to removing the code entirely) basically as part of my own workflow process. I have various `grep` aliases that look for different types of markup that I may put into a code - block comments, fixmes, todos, etc. In all cases I'm trying to draw attention to something that I want to address at some point (but not at the moment, because I'm working on other things), and generally I'll only leave an `#if 0` sitting around if it's something I'm not sure I want to keep, or remove. In this particular instance, I have a fixme right above it, so I suppose I could just change it to a block comment if `#if 0` bothers anyone; either way, I'll find it. Speaking of the various `write()` functions, going through the entire code (which has grown considerably since I wrote the first `write()s`), and ensuring that the various flavors of `write()` are consistent with one another and reasonably clear to potential users, is relatively high on my priority list. So, I'll probably be looking at this exact code sometime soon. At that point, I'll either put the function in question back in, or remove it entirely.

src/GNDStk/JSON/src/write.hpp
staleyLANL: No worries. I actually use unnamed parameters all the time, when I need the function signature to be a particular thing, but *know* that I'm never going to use the parameter. When I do suspect that I'll use the parameter at some point, I'll generally do the `(void)` thing, precisely because I can easily `grep` for `(void)` and identify such things, when I've dealt with other priorities and am ready to deal with them. (See my comment about the `#if 0` business and `grep`; the same principle is at work here.)

python/src/v1.9/transport/ReactionSuite.python.cpp
staleyLANL: It looks like this is autogenerated code. So, the question is why the code generator doesn't produce that code block any longer.

staleyLANL: So, I haven't, as yet, been able to figure out what's going on here. I'm not seeing the phrase `def_static` in the code generator in either the develop or master branch. I don't typically mess with the python code generation, and have no recollection of doing anything that would have caused the above code block to disappear. Looking back at the master branch, I see the above construction in ReactionSuite.python.cpp, but not in AnythingElse.python.cpp. Was it possibly from some sort of customization you wrote?? If so, then perhaps the customization inadvertently got blown away by some process.

autogen/.gitignore
whaeck: Ah, I see. You should not make changes related to one pull request in another one. I'll just remove the file and then the change will just propagate forward.

python/src/v1.9/transport/ReactionSuite.python.cpp
whaeck: It was customisation on my part. I'll just put it back with a better customisation for python. I think we can do something a bit more robust. I'll branch off this branch and make a pull request to pull it into this branch before we merge.

src/GNDStk/HDF5.hpp
```diff
@@ -0,0 +1,126 @@
...
+   // temporaryName()
+   static std::string temporaryName()
+   {
+      // Generate a file name that's supposed to be suitable for safely creating
+      // a temporary file. L_tmpnam and tmpnam() are from <cstdio>. In case
+      // anyone wonders, we don't need a +1 in the [L_tmpnam]. :-)
+      static char buffer[L_tmpnam];
+
+      while (true) {
+         const std::string name = std::string(tmpnam(buffer)) + ".h5";
```
whaeck, line 32: I am getting tmpnam is deprecated warning on clang 12. Due to us handling every warning as an error, it does not compile onn clang 12. We should replace this asap.

staleyLANL: This was long since replaced with better code in later PRs, which no longer get that warning.

python/src/v1.9/transport/ReactionSuite.python.cpp
staleyLANL: By the way, I saw this only for ReactionSuite, which of course is the "top level" node in our prototype hierarchy. I see no reason that a similar capability couldn't be done for all other generated (read: derived from class Component) classes.

src/GNDStk/HDF5.hpp
whaeck: Good. I switched to g++ 9.3 on my laptop so I can actually compile it now.

staleyLANL: By the way, it's worth explaining why I deal with temporary files when I'm working with HDF5.

For ensuring consistency of behavior, and making various things throughout GNDStk (reads, writes; XML, JSON, and HDF5; streams and files; convert functions; etc.) work smoothly and consistently with one another, I have lightweight wrappers around pugixml's data structure for XML, nlohmann json's data structure for JSON, and HighFive's data structure for HDF5. (Which, itself, actually relies on a large and complex underlying C-based library written by the official HDF5 people.)

When XML is read by pugixml, the whole file is put into pugxmli's internal data structure. So far so good.

When JSON is read by nlohmann json, the whole file is put into nlohmann json's internal data structure. So far so good.

Oops...HighFive (probably because of the underlying C-based library) is different!! Its data structure is small, and mostly serves as essentially a *file* *descriptor* that points to a file on disk! HDF5 data aren't actually stored in memory. :-(

That throws a monkey wrench into the whole system. So, of course, wanting GNDStk to be as awesome a code as possible, my HDF5 wrapper un-monkey-wrenches things.

Imagine a simple implementation that was unaware of what the underlying hdf5-related libraries actually do. You might "load" an HDF5 file into an object of my HDF5 class, then copy that to a second HDF5 object that you wish to modify. Unfortunately, you'd basically be dealing with a shallow copy - two things that amount to file descriptors to the same underlying file. You can imagine various bad things that would happen.

So, I wrote all of my HDF5 capabilities to be fully aware of the issue. Just say - to look at a simple case - that you do what I describe above: read into an object of class HDF5, then copy to another. For proper semantics with no surprises, GNDStk's copy capability *duplicates* the original file (for which the source HDF5 object has a file descriptor) into a temporary file. (Hence the need to generate temporary file names!) Then, the copy's destination HDF5 is set up to have a file descriptor to the temporary file, not to the original file.

I carefully wrote all the behavior of all of GNDStk's HDF5 I/O to deal with this, so that a high-level user will just see an interface that looks exactly the same as the interfaces for XML and JSON. This was a tricky task, but one that I think the best GNDS code in the world deserves!!

Be aware, then, that if you happen to look in detail at everything I did for HDF5, you'll see various *apparently* funny gymnastics and convolutednesses (if that's a word) that might seem bizarre and unnecessary, unless you understand that I'm dealing with underlying code that doesn't (or only partially) has actual HDF5 data structures in memory.

PR #70, opened 2022-01-19
Feature/data truncate

staleyLANL:
See commit messages. It looks like I hadn't put "remove unused built-in Meta and Child objects" in as its own PR back several days ago after I did it. I'd intended to, but that's OK, it can go here. The main point of this PR was putting in a feature that allows partial printing of (possibly very large) data sets in BlockData objects. Of course I updated the relevant test code to include tests of the new feature as well.


staleyLANL:
> See commit messages. It looks like I hadn't put "remove unused built-in Meta and Child objects" in as its own PR back several days ago after I did it. I'd intended to, but that's OK, it can go here. The main point of this PR was putting in a feature that allows partial printing of (possibly very large) data sets in BlockData objects. Of course I updated the relevant test code to include tests of the new feature as well.

Oops, my original comment needs fixing. I *had* put the removal of unused `Meta` and `Child` objects in another PR, and just hadn't specified the correct base for this new PR. That should all be updated now.

PR #71, opened 2022-01-20
Simple commit: limit lines to 80 characters in many places.

staleyLANL:
We're generally trying to limit source-code lines to 80 characters or less in length. This isn't an absolute requirement, but, in keeping with the idea, I reformatted longer lines here and there in the hpp and cpp files throughout the code base, where doing so made sense.

Sometimes, such codes have string literals (as with `R"..."`) with lines of more than 80 characters. Those I left as-is; the lines in question are supposed to be precisely the way they are.

Also, for now, I didn't reformat long lines in other types of files. That includes, for example, various `CMakeLists.txt` files throughout GNDStk.

Building very slightly on an earlier PR, I also made a small change to the `GNDStk::columns` variable, so that it would be more consistent, in terms of usage, with the related `GNDStk::truncate` variable. This has nothing to do with 80+ character lines, but was a simple change, so I decided to include it here.

As usual, I modified a few comments and made small cosmetic changes in a few places.


whaeck:
I don't see why we suddenly need to start enforcing the 80 character rule. As you know, going over 80 characters has never disturbed me. Since this is mainly a cosmetic change, I'll just merge this in its daughter branch.

staleyLANL:
> I don't see why we suddenly need to start enforcing the 80 character rule. As you know, going over 80 characters has never disturbed me. Since this is mainly a cosmetic change, I'll just merge this in its daughter branch.

Yes, it's not really a big deal either way, and it's impractical to make everything obey an official "<80 lines" rule. I think I'd done some much more difficult things before this, and in part just wanted something simple to work on for a bit. And, I think that such a rule does help maintain some discipline in a code, and it's certainly good if somebody wanted to print a code on paper, where about 80 lines is typically about what prints in a typical font before line wrapping.

